PR #13, opened 2023-05-01
Bring elemental-cli code into the agent

Itxaka:
Fixes https://github.com/kairos-io/kairos/issues/1210


Itxaka:
only install for now uses elemental libs directly integrated, seems to work as expected.
upgrade and reset will come later.

config merge will be a pita! elemental does a lot of stuff...

And still missing dropping all unneeded stuff for build...

Itxaka:
upgrade+reset done

reset fails with a panic so not ready yet :D

Itxaka:
Reset Works! Upgrade Works!

internal/agent/install.go
```diff
@@ -249,47 +251,29 @@ func Install(dir ...string) error {
 }
 
 func RunInstall(options map[string]string) error {
+	// TODO: run yip directly
 	utils.SH("elemental run-stage kairos-install.pre")             //nolint:errcheck
```
jimmykarily, line 255: I thought the goal of this PR was to not have "elemental" in the image at all. Are we doing it in steps?

internal/agent/install.go
```diff
@@ -298,16 +282,56 @@ func RunInstall(options map[string]string) error {
...
+	// Make it configurable?
+	installConfig, err := elementalConfig.ReadConfigRun("/etc/elemental")
+	if err != nil {
+		return err
+	}
+
+	// Set our cloud-init to the file we just created
+
+	_, reboot := options["reboot"]
+	_, poweroff := options["poweroff"]
```
jimmykarily, line 297:
```suggestion
	reboot := options["reboot"]
	poweroff := options["poweroff"]
```

internal/agent/install.go
```diff
@@ -298,16 +282,56 @@ func RunInstall(options map[string]string) error {
...
-	if err := cmd.Run(); err != nil {
+	// Load the installation Config from the system
+	// TODO: This uses the default mounter, logger, fs, etc...
+	// Make it configurable?
+	installConfig, err := elementalConfig.ReadConfigRun("/etc/elemental")
+	if err != nil {
+		return err
+	}
+
+	// Set our cloud-init to the file we just created
```
jimmykarily, line 294: This comment belongs above line 306 below I think

internal/agent/reset.go
```diff
@@ -14,27 +13,32 @@ import (
...
 func Reset(dir ...string) error {
+	// TODO: Enable args? No args for now so no possibility of reset persistent or overriding the source for the reset
+	// Nor the auto-reboot via cmd?
 	bus.Manager.Initialize()
 
-	options := map[string]string{}
-
 	bus.Manager.Response(sdk.EventBeforeReset, func(p *pluggable.Plugin, r *pluggable.EventResponse) {
-		err := json.Unmarshal([]byte(r.Data), &options)
+		err := json.Unmarshal([]byte(r.Data), &map[string]string{})
```
jimmykarily, line 31: We used to unmarshal to `options` because we needed that variable later. What's the reason of unmarshaling now? Just to print the error if it's not "umarshable"?

internal/agent/reset.go
```diff
@@ -14,27 +13,32 @@ import (
...
 	"github.com/kairos-io/kairos/v2/pkg/config/collector"
+	"github.com/kairos-io/kairos/v2/pkg/elementalConfig"
 
 	"github.com/mudler/go-pluggable"
 	"github.com/pterm/pterm"
 )
 
 func Reset(dir ...string) error {
+	// TODO: Enable args? No args for now so no possibility of reset persistent or overriding the source for the reset
+	// Nor the auto-reboot via cmd?
```
jimmykarily, line 27: afaict `--reset-persistent` flag was set through `r.Data` in this code. Why can't we keep it this way?

internal/agent/upgrade.go
```diff
@@ -117,24 +121,51 @@ func Upgrade(
...
+		Username:      authUser,
+		Password:      authPass,
+		ServerAddress: authServer,
+		Auth:          authType,
+		IdentityToken: identityToken,
+		RegistryToken: registryToken,
+	}
+
+	// Override the default luet to pass the auth
+	// Remember to create the temp dir!
```
jimmykarily, line 144: Is this a TODO note? Who should remember it and when?

internal/agent/upgrade.go
```diff
@@ -117,24 +121,51 @@ func Upgrade(
...
+	tmpDir := elementalUtils.GetTempDir(&resetConfig.Config, "")
+	l := luet.NewLuet(luet.WithLogger(resetConfig.Logger), luet.WithAuth(auth), luet.WithLuetTempDir(tmpDir))
+	resetConfig.Luet = l
+
+	// Generate the upgrade spec
+	resetSpec, err := elementalConfig.ReadUpgradeSpec(resetConfig)
```
jimmykarily, line 150: This variable should probably be called `upgradeSpec`

internal/agent/upgrade.go
```diff
@@ -117,24 +121,51 @@ func Upgrade(
...
-		"--auth-username", authUser,
-		"--auth-password", authPass,
-		"--auth-server-address", authServer,
-		"--auth-type", authType,
-		"--auth-registry-token", registryToken,
-		"--auth-identity-token", identityToken,
-	)
-
+	// Load the upgrade Config from the system
+	resetConfig, err := elementalConfig.ReadConfigRun("/etc/elemental")
```
jimmykarily, line 125: This variable should probably be called `upgradeConfig` (though it seems that we load the same config on reset so maybe a more generic name in both cases would be better)

pkg/action/build-iso.go
```diff
@@ -0,0 +1,287 @@
+/*
```
jimmykarily, line 1: I don't think we should bring the build parts too. There is another story to move those to osbuilder (https://github.com/kairos-io/kairos/issues/711). Until we do, I guess it's ok to keep them in the old "elemental-cli" binary, what do you think? Or are we planning to replace elemental-cli with kairos-agent first and then move the build parts to osbuilder?

internal/agent/install.go
Itxaka: this is a map[string]string], so cant get just that without more work afterwards. Doing it trhis way means that the reboot and poweroff are directly a bool type :)

internal/agent/reset.go
Itxaka: options was an empty `map[string]string{}` before. This is just removing an extra var. No idea why it was done like this before, I guess because all install,reset,upgrade do the same? In case we add options later?

internal/agent/reset.go
Itxaka: This comment relates for the cli being called. If you call kairos-agent reset  there is no flags that can be passed manually to override the default config. Im just thinking that maybe we should be able to override the config via flags.

internal/agent/upgrade.go
Itxaka: Well this has history :D
Its just a note to mention that when creating a custom luet object you need to create a temp dir for it explicitely. Otherwise things go wrong very soon. I'll expand the comment a bit more!

internal/agent/upgrade.go
Itxaka: wops! good catch!

internal/agent/upgrade.go
Itxaka: indeed. It actually should be config but there is already like 32 different config vars lol. I'll change it!

pkg/action/build-iso.go
Itxaka: yep yep, this is a "fast, dirty" job to bring the libs in here, there is still lot of work to clean this up and unify, but I didnt want to block this further. Removing those parts can be done afterwards in parallel with the config merge and others.

internal/agent/install.go
Itxaka: we can work on that afterwards in paralell with other work, no need to block this further with that

internal/agent/reset.go
Itxaka: oooh I see, it comes from the `r.Data`.

Where the fuck does that r.Data magically appears from?

Itxaka: I have no idea where that comes from or how does it get set to be honest, I'll bring it up on the meeting today.

Also good catch, this was setting reset-persistent by default in no data was coming in, I missed that!

internal/agent/install.go
jimmykarily: could `options["reboot"]` be set to `"false"` ? Because in that case `reboot` would be `true` (because the key exists), although it's value is "false"

internal/agent/upgrade.go
jimmykarily: But there is nothing to do in this part of the code no? Maybe this reminder belongs closer to where the temp dir should be created.

internal/agent/upgrade.go
jimmykarily: Yes, we will hopefully consolidate configs [here](https://github.com/kairos-io/kairos/issues/1225) (That story is a can of worms...)

internal/agent/upgrade.go
Itxaka: It should be created here, when you create the new luet. I hope to fix this in the future once the code is folder but currently if someone removes that line, unintended consequences will happen :D

internal/agent/install.go
Itxaka: That is a good point!

Itxaka: welp, this actually does the same behaviour that the old one! so yeah, not touching this in the PR, maybe afterwards as its a behaviour change!

jimmykarily (approved):
I'm fine to merge this as it is and improve the code in future PRs. After all, this is mainly a code migration from another repository. No need to fix everything at once.

PR #17, opened 2023-05-09
Add missing patch

Itxaka:
Patch was missing when extracting the agent into its own repo


mauromorales (approved):
lgtm, mentioned this PR on the card about the refactoring so we don't forget to remove from agent when extracting to sdk

PR #18, opened 2023-05-09
Mount OEM during kcrypt hook

Itxaka:
We need to read the config and its stored there

### DIFF
--- a/internal/agent/hooks/kcrypt.go
+++ b/internal/agent/hooks/kcrypt.go
@@ -2,6 +2,7 @@ package hook
 
 import (
 	"fmt"
+	"github.com/kairos-io/kairos-sdk/machine"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/agent/hooks/kcrypt.go
```diff
@@ -16,6 +17,12 @@ func (k Kcrypt) Run(c config.Config) error {
 		return nil
 	}
 
+	// Config passed during install ends up here, so we need to read it
+	_ = machine.Mount("COS_OEM", "/oem")
+	defer func() {
+		_ = machine.Umount("/oem") //nolint:errcheck
+	}()
```
jimmykarily, line 24: So much for encrypting OEM I guess... :D Unlocking needs the config too right? (to find the challenger server url etc).

PR #19, opened 2023-05-11
Allow overriding the schem version number

Itxaka:
We currently default to the agent number but there could be reason why we want this being override. For example we dont really sync with the kairos OS release schedule so there is no way we can sync up with that.

This allows to override the version number so systems can generate a schema that is synced to their version number for ease of use.


mauromorales (approved):
lgtm

PR #22, opened 2023-05-16
:robot: Use cypress container for webui tests

Itxaka:
It speeds up the tests a lot as it already has all the deps included


mauromorales (approved):
nice finding :sunglasses:

PR #24, opened 2023-05-17
Bump sdk deps and adjust image extraction

Itxaka:
image extraction has moved to sdk as its being reused in different places. Use that directly.


mauromorales (approved):
lgtm

PR #27, opened 2023-05-17
:sparkles: Allow adding extra cloud-init paths to run-stage

Itxaka:
This was available on elemental but not on agent. And its very handy :D


mudler (approved):
Looks good!

PR #28, opened 2023-05-18
Add root_path flag to install-bundle

Itxaka:
There is no way of setting the default root path if you manually install a bundle, so expose that as a flag in the cli

### DIFF
--- a/main.go
+++ b/main.go
@@ -197,6 +197,10 @@ E.g. kairos-agent install-bundle container:quay.io/kairos/kairos...
 				EnvVars: []string{"REPOSITORY"},
 				Value:   "docker://quay.io/kairos/packages",
 			},
+			&cli.StringFlag{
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -197,6 +197,10 @@ E.g. kairos-agent install-bundle container:quay.io/kairos/kairos...
 				EnvVars: []string{"REPOSITORY"},
 				Value:   "docker://quay.io/kairos/packages",
 			},
+			&cli.StringFlag{
+				Name:  "root_path",
```
jimmykarily, line 201: we used dashes `-` for `local-file` below. Maybe keep it consistent?

jimmykarily (approved):

main.go
Itxaka: good catch!

PR #31, opened 2023-05-24
Add debug flag in root cmd

Itxaka:
This is now passed to elemental actions so they can put the debug output properly

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/sirupsen/logrus"
 	"net/url"
 	"os"
 	"strings"
@@ -74,7 +75,7 @@ func mergeOption(cloudConfig string, r map[string]string) {
 	}
 }
 
-func ManualInstall(c string, options map[string]string, strictValidations bool) error {
+func ManualInstall(c string, options map[string]string, strictValidations, debug bool) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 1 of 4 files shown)


mauromorales (approved):
lgtm

main.go
```diff
@@ -588,6 +591,11 @@ func main() {
 				Usage:   "Fail instead of warn on validation errors.",
 				EnvVars: []string{"STRICT_VALIDATIONS"},
 			},
+			&cli.BoolFlag{
+				Name:    "debug",
+				Usage:   "enable debug output",
+				EnvVars: []string{"KAIROS_AGENT_DEBUG"},
+			},
```
jimmykarily, line 598: don't we need to add a similar block to `manual-install`, `reset` etc?

internal/agent/install.go
```diff
@@ -98,13 +99,17 @@ func ManualInstall(c string, options map[string]string, strictValidations bool)
 	mergeOption(configStr, options)
 
 	if options["device"] == "" {
-		options["device"] = cc.Install.Device
+		if cc.Install.Device == "" {
+			options["device"] = detectDevice()
```
jimmykarily, line 103: is this enough to allow us to run `kairos-agent manual-install config.yaml` (without specifying `--device`)?
I remember there is a check the makes the command fail when `--device` is not specified.

Itxaka: it actually panics lol

--device is NOT a required flag, so you can pass an empty value and call it a day. This may have changed behaviour a bit, but currently is broken.

main.go
Itxaka: nope, this is a global flag at the main command level, so its available to all commands

PR #32, opened 2023-05-25
1225 config collector elemental

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/1225

and probably https://github.com/kairos-io/kairos/issues/209 too

There are basically 3 fixes here:

- Merging the user's cloud config before we run the before-install and `kairos-install.pre.*` hooks
- Creating all partitions defined in the cloud config, not just the first
- Expanding the last partition after we have created all the defined partitions, so that we expand the partition we defined.

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -5,12 +5,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/sirupsen/logrus"
 	"net/url"
 	"os"
 	"strings"
 	"syscall"
 	"time"
+
+	"github.com/sirupsen/logrus"
 
 	events "github.com/kairos-io/kairos-sdk/bus"
 	"github.com/kairos-io/kairos-sdk/machine"
@@ -29,18 +30,6 @@ import (
 	"github.com/pterm/pterm"
 	"gopkg.in/yaml.v2"
 )
-
-func optsToArgs(options map[string]string) (res []string) {
-	for k, v := range options {
-		if k != "device" && k != "cc" && k != "reboot" && k != "poweroff" {
-			res = append(res, fmt.Sprintf("--%s", k))
-			if v != "" {
-				res = append(res, v)
-			}
-		}
-	}
-	return
-}
 
 func displayInfo(agentConfig *Config) {
 	fmt.Println("--------------------------")
@@ -106,7 +95,14 @@ func ManualInstall(c string, options map[string]string, strictValidations, debug
 		}
 	}
 
-	return RunInstall(debug, options)
+	// Load the installation Config from the system
+	installConfig, err := elementalConfig.ReadConfigRun("/etc/elemental")
+	if err != nil {
+		return err
+	}
+	installConfig.Debug = debug
+
+	return RunInstall(installConfig, options)
 }
 
 func Install(debug bool, dir ...string) error {
@@ -133,6 +129,13 @@ func Install(debug bool, dir ...string) error {
 
 	ensureDataSourceReady()
 
+	// Load the installation Config from the system
+	installConfig, err := elementalConfig.ReadConfigRun("/etc/elemental")
+	if err != nil {
+		return err
+	}
+	installConfig.Debug = debug
+
 	// Reads config, and if present and offline is defined,
 	// runs the installation
 	cc, err := config.Scan(collector.Directories(dir...), collector.MergeBootLine, collector.NoLogs)
@@ -145,7 +148,7 @@ func Install(debug bool, dir ...string) error {
 		r["device"] = cc.Install.Device
 		mergeOption(configStr, r)
 
-		err = RunInstall(debug, r)
+		err = RunInstall(installConfig, r)
 		if err != nil {
 			return err
 		}
@@ -239,7 +242,7 @@ func Install(debug bool, dir ...string) error {
 
 	pterm.Info.Println("Starting installation")
 
-	if err := RunInstall(debug, r); err != nil {
+	if err := RunInstall(installConfig, r); err != nil {
 		return err
 	}
 
@@ -256,19 +259,10 @@ func Install(debug bool, dir ...string) error {
 	return nil
 }
 
-func RunInstall(debug bool, options map[string]string) error {
-	// Load the installation Config from the system
-	installConfig, err := elementalConfig.ReadConfigRun("/etc/elemental")
-	if err != nil {
-		return err
-	}
-	if debug {
+func RunInstall(installConfig *v1.RunConfig, options map[string]string) error {
+	if installConfig.Debug {
 		installConfig.Logger.SetLevel(logrus.DebugLevel)
 	}
-
-	// Run pre-install stage
-	_ = elementalUtils.RunStage(&installConfig.Config, "kairos-install.pre", installConfig.Strict, installConfig.CloudInitPaths...)
-	events.RunHookScript("/usr/bin/kairos-agent.install.pre.hook") //nolint:errcheck
 
 	f, _ := os.CreateTemp("", "xxxx")
 	defer os.RemoveAll(f.Name())
@@ -292,7 +286,7 @@ func RunInstall(debug bool, options map[string]string) error {
 	env := append(c.Install.Env, c.Env...)
 	utils.SetEnv(env)
 
-	err = os.WriteFile(f.Name(), []byte(cloudInit), os.ModePerm)
+	err := os.WriteFile(f.Name(), []byte(cloudInit), os.ModePerm)
 	if err != nil {
 		fmt.Printf("could not write cloud init: %s\n", err.Error())
 		os.Exit(1)
@@ -312,6 +306,7 @@ func RunInstall(debug bool, options map[string]string) error {
 	if err != nil {
 		return err
 	}
+
 	// Set our cloud-init to the file we just created
 	installSpec.CloudInit = append(installSpec.CloudInit, f.Name())
 	// Get the source of the installation if we are overriding it
@@ -338,6 +333,15 @@ func RunInstall(debug bool, options map[string]string) error {
 	if err != nil {
 		return err
 	}
+
+	// TODO: This should be enough to make the whole user's config available to
+	// "elemental". Thus the before-install hook should also be available here:
+	// https://github.com/kairos-io/kairos/issues/209#issuecomment-1457938851
+	installConfig.CloudInitPaths = append(installConfig.CloudInitPaths, installSpec.CloudInit...)
+	// Run pre-install stage
+	_ = elementalUtils.RunStage(&installConfig.Config, "kairos-install.pre", installConfig.Strict, installConfig.CloudInitPaths...)
+	events.RunHookScript("/usr/bin/kairos-agent.install.pre.hook") //nolint:errcheck
+
 	// Create the action
 	installAction := action.NewInstallAction(installConfig, installSpec)
 	// Run it
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/sanity-io/litter"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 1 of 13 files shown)


internal/agent/install.go
```diff
@@ -106,7 +95,14 @@ func ManualInstall(c string, options map[string]string, strictValidations, debug
 		}
 	}
 
-	return RunInstall(debug, options)
+	// Load the installation Config from the system
+	installConfig, err := elementalConfig.ReadConfigRun("/etc/elemental")
```
jimmykarily, line 99: DRY this (an put the path to a constant)

Itxaka (commented):
I was expecting this to be different.

i.e. the ReadConfigRun to call the collector so we merge options from elemental config <- cloud-config so we can move the options out of elemental, so they are configured during install/upgrade/reset directly? Otherwise this still reads most of stuff from elemental config which is on RO and not easily modificable and we need to get exposing things over to the agent.

Ideally IMO, one function should read both files and merge them, and we can deprecate the elemental config directly and have everything by default unless set on the cloud-config.

thoughts?

jimmykarily:
> I was expecting this to be different.
> 
> i.e. the ReadConfigRun to call the collector so we merge options from elemental config <- cloud-config so we can move the options out of elemental, so they are configured during install/upgrade/reset directly? Otherwise this still reads most of stuff from elemental config which is on RO and not easily modificable and we need to get exposing things over to the agent.
> 
> Ideally IMO, one function should read both files and merge them, and we can deprecate the elemental config directly and have everything by default unless set on the cloud-config.
> 
> thoughts?

I was trying to understand the various different configs and this is what I ended up with:

- [pkg/config/config.go](https://github.com/kairos-io/kairos-agent/blob/5fe8eb8cc2617e96e609da282d1079a95ba06e0c/pkg/config/config.go#L38) is the struct that holds only the relevant parts from the yaml collected with the collector. We have similar structs in the other consumers of the collector. This is the struct to which we unmarshal the yaml created by the collector. This is where user's config ends up being unmarshalled to.
- [pkg/types/v1/config.go](https://github.com/kairos-io/kairos-agent/blob/5fe8eb8cc2617e96e609da282d1079a95ba06e0c/pkg/types/v1/config.go#L42) seems to be a relic wrapper from when we used to call out to `elemental` (wrapping the command line options?). We should get rid of this by moving the various options where they make more sense (other configs)
- `RunConfig` : yet another "config" which also embeds the one above. This can be gone too?
- `InstallSpec` : doesn't have a "config" in its name, but it's still another "config". 

I tried to clean my head from what I've seen and think about what kind of configuration we actually need. I see 2 different types:

- The cloud-config including the user provided one
- The cli configuration which is simply [optionally] replacing command line arguments that can be passed to the binary.

The user should be able to change both (no hardcoded locations, but maybe default config for the cli could exist).

If this is a good plan ^ I would need some help consolidating and cleaning up because I'm not sure what all those options are used for.

Still, we can still move on with this PR as a fix for custom partitioning and refactor on another PR.

Itxaka:
> > I was expecting this to be different.
> > i.e. the ReadConfigRun to call the collector so we merge options from elemental config <- cloud-config so we can move the options out of elemental, so they are configured during install/upgrade/reset directly? Otherwise this still reads most of stuff from elemental config which is on RO and not easily modificable and we need to get exposing things over to the agent.
> > Ideally IMO, one function should read both files and merge them, and we can deprecate the elemental config directly and have everything by default unless set on the cloud-config.
> > thoughts?
> 
> I was trying to understand the various different configs and this is what I ended up with:
> 
>     * [pkg/config/config.go](https://github.com/kairos-io/kairos-agent/blob/5fe8eb8cc2617e96e609da282d1079a95ba06e0c/pkg/config/config.go#L38) is the struct that holds only the relevant parts from the yaml collected with the collector. We have similar structs in the other consumers of the collector. This is the struct to which we unmarshal the yaml created by the collector. This is where user's config ends up being unmarshalled to.
> 

This was the config from the kairos-agent only, and it used parts of that to build the args to pass to elemental. This was loaded from the cloud-config/config or interactive install inputs

>     * [pkg/types/v1/config.go](https://github.com/kairos-io/kairos-agent/blob/5fe8eb8cc2617e96e609da282d1079a95ba06e0c/pkg/types/v1/config.go#L42) seems to be a relic wrapper from when we used to call out to `elemental` (wrapping the command line options?). We should get rid of this by moving the various options where they make more sense (other configs)
> 

This is the main config of elemental. This is what is used around in a lot of different places and exposes *interfaces* so they can be easily substituted by other structs that can fulfill that interface, see for example the `Mounter` interface and how we have a mocked one so we can tests the whole thing without resorting to e2e tests.

This would be impossible to remove as its being used almost everywhere for everything. You need to log? cfg.Logger, mount stuff? cfg.Mounter. 

Not only that but allows for future proofing without too much hassle. Want to substitute the current mounter? Just write a new one or adapt an existing lib to fullfill the interface. Done.

>     * `RunConfig` : yet another "config" which also embeds the one above. This can be gone too?

Runconfig was the config that affected run stuff (install,reset,upgrade). It embeds the config on it and it was especialized for that run config with values that were only used for it. There was a BuildConfig that had different values that was dedicated to building stuff as as separation of concerns. IMO this one could go away easily as we only have one config there for install/upgrade/reset and the build stuff is somewhere else
> 
>     * `InstallSpec` : doesn't have a "config" in its name, but it's still another "config".
> 
> 


InstallSpec/UpgradeSpec/ResetSpec  is the final spec to be used for action X. Its the results of reading the config and sanitizing it. Its the guide to the action and the values that will be used. Its a byproduct of the config + sanitizing it + adding defaults + runtime stuff (like platform arch).


> I tried to clean my head from what I've seen and think about what kind of configuration we actually need. I see 2 different types:
> 
>     * The cloud-config including the user provided one
> 
>     * The cli configuration which is simply [optionally] replacing command line arguments that can be passed to the binary.
> 
> 
> The user should be able to change both (no hardcoded locations, but maybe default config for the cli could exist).
> 
> If this is a good plan ^ I would need some help consolidating and cleaning up because I'm not sure what all those options are used for.
> 
> Still, we can still move on with this PR as a fix for custom partitioning and refactor on another PR.


IMO ideally we should.

 - Drop the RunConfig and use a expanded Config. 
 - Move the loading of the values from /etc/elemental/config.yaml to the cloud-config loader. Deprecate loading from /etc/elemental/config.yaml
 - Load from the collector the values and dump them into a Config object (expanded)
 - Use that Config everywhere. Rely on it for all the stuff used in the lifecycle of the agent (so log with cfg.Logger for example)
 - On Config.Sanitize() or init or whatever, add the default values to empty values.



So basically, your cloud config would for example look like a merge of the elemental config and our current cloud-config:

```
install:
  auto: true
  reboot: true
  device: /dev/vda
  grub-entry-name: "Kairos"
  system:
    size: 3000
  grub_options:
    extra_cmdline: "rd.immucore.debug foobarzz"

stages:
  initramfs:
    - name: "Set user and password"
      users:
        kairos:
          passwd: "kairos"
      hostname: kairos-{{ trunc 4 .Random }}

```

And that would generate a Config object that had `Config.Auto: true`, `Config.Reboot: true`, etc...

pkg/cloudinit/layout_plugin.go
```diff
@@ -111,7 +102,16 @@ func layoutPlugin(l logger.Interface, s schema.Stage, fs vfs.FS, console plugins
...
 	}
+
+	if s.Layout.Expand != nil {
+		l.Infof("Extending last partition up to %d MiB", s.Layout.Expand.Size)
+		out, err := dev.ExpandLastPartition(s.Layout.Expand.Size)
+		if err != nil {
+			l.Error(out)
+			return err
+		}
+	}
```
jimmykarily, line 114: NOTE: Let's try an installation without custom partitions with this fix to see if this breaks something. @Itxaka has a hunch...

jimmykarily:
I run a `kairos-agent manual-install` with now custom partitions,  using a custom build of kairos-agent from this branch and I got this error:

```
INFO[2023-05-29T15:12:06Z] Copying /run/cos/state/cOS/active.img source... 
INFO[2023-05-29T15:12:09Z] Finished copying /run/cos/state/cOS/active.img into /run/cos/state/cOS/passive.img 
INFO[2023-05-29T15:12:10Z] Running after-install hook                   
ERRO[2023-05-29T15:12:10Z] mount: /tmp/mnt/STATE: /dev/vda4 already mounted on /run/cos/state.
 : failed to run STATEDIR=/tmp/mnt/STATE
STATE=$(blkid -L COS_STATE || true)
mkdir -p $STATEDIR || true
mount ${STATE} $STATEDIR
: exit status 32 
ERRO[2023-05-29T15:12:10Z] 1 error occurred:
	* failed to run STATEDIR=/tmp/mnt/STATE
STATE=$(blkid -L COS_STATE || true)
mkdir -p $STATEDIR || true
mount ${STATE} $STATEDIR
: exit status 32
 
ERRO[2023-05-29T15:12:10Z] umount: /tmp/mnt/STATE: not mounted.
 : failed to run umount /tmp/mnt/STATE
: exit status 32 
ERRO[2023-05-29T15:12:10Z] 1 error occurred:
	* failed to run umount /tmp/mnt/STATE
: exit status 32
 
INFO[2023-05-29T15:12:10Z] Unmounting disk partitions                   
INFO[2023-05-29T15:12:10Z] kairos-agent version v2.1.0-rc4-3-g1a2174d-dirty 
INFO[2023-05-29T15:12:10Z] Running stage: kairos-install.after.before   
INFO[2023-05-29T15:12:10Z] Done executing stage 'kairos-install.after.before' 
INFO[2023-05-29T15:12:10Z] Running stage: kairos-install.after          
INFO[2023-05-29T15:12:10Z] Done executing stage 'kairos-install.after'  
INFO[2023-05-29T15:12:10Z] Running stage: kairos-install.after.after    
INFO[2023-05-29T15:12:10Z] Done executing stage 'kairos-install.after.after' 
INFO[2023-05-29T15:12:10Z] Running stage: kairos-install.after.before   
INFO[2023-05-29T15:12:10Z] Done executing stage 'kairos-install.after.before' 
INFO[2023-05-29T15:12:10Z] Running stage: kairos-install.after          
INFO[2023-05-29T15:12:10Z] Done executing stage 'kairos-install.after'  
INFO[2023-05-29T15:12:10Z] Running stage: kairos-install.after.after    
INFO[2023-05-29T15:12:10Z] Done executing stage 'kairos-install.after.after' 
[   89.083940][ T1778] EXT4-fs (vda5): mounted filesystem with ordered data mode. Opts: (null). Quota mode: none.
[   89.090567][ T1783] EXT4-fs (vda2): mounted filesystem with ordered data mode. Opts: (null). Quota mode: none.
```

@Itxaka is this relevant to what you were afraid would happen? Does it ring any bells?

Itxaka:
Not that I can remember. That looks like the boot assessment after-install run

jimmykarily:
hmm I tried again with [the full config](https://github.com/kairos-io/kairos/issues/209#issuecomment-1564357778) and I get the same error when I use custom partitions:

```
ERRO[2023-05-29T15:21:31Z] umount: /tmp/mnt/STATE: not mounted.
 : failed to run umount /tmp/mnt/STATE
: exit status 32 
ERRO[2023-05-29T15:21:31Z] 1 error occurred:
	* failed to run umount /tmp/mnt/STATE
: exit status 32
 
INFO[2023-05-29T15:21:31Z] Unmounting disk partitions   
```

I don't know how this error popped up. It wasn't there on Friday, I'm pretty sure.

jimmykarily:
Ignore it all, I thought any "recent" version of kairos core would be ok and used `kairos-opensuse-leap-v2.0.2-k3sv1.23.17+k3s1.iso`. Turns out it works fine on a build from latest master, both with and without custom partitions. I don't get any errors.

jimmykarily:
We created most of the needed tests. The only one missing I think is the one that tests this fix: https://github.com/kairos-io/kairos-agent/pull/32/files#diff-bdceadb6b1ed909e97670e80e5251016ad0826cff05ec4cac9ed83add43cd992R338-R345

which ensures that the users cloud config is merge before we run the hooks. I started writing a test here: https://github.com/kairos-io/kairos-agent/pull/32/files#diff-f26da7c79a4ee9afdb26e2f102d65f35feea2f2d731b3d1bdd4d48349c88e0ebR118 but I find that impossible to mock enough things to get `RunInstall` to work in tests. It a high level function and doesn't take interface arguments like others so it's not "mockable" in its current form.

pkg/config/config.go
```diff
@@ -23,6 +23,7 @@ const (
 type Install struct {
 	Auto                   bool              `yaml:"auto,omitempty"`
 	Reboot                 bool              `yaml:"reboot,omitempty"`
+	NoFormat               bool              `yaml:"no-format,omitempty"`
```
mauromorales, line 26: All other are using underscores ... not a biggie if this is like this for another reason

```suggestion
	NoFormat               bool              `yaml:"no_format,omitempty"`
```

mauromorales (approved):
lgtm

PR #33, opened 2023-05-26
Run the user in initramfs if we dont need network

Itxaka:
If the user does not select sshkeys or selects the default values during interactive install, there is no need to run the users in the network stage.

That causes the machine to take a while to create the users when booting which causes to not be able to log in until the network stage is done.

This patch makes it so it cleans up the default non-valid users value and set the stage to initramfs so login to the machine is inmmediate.


mudler (approved):
nice! like also the cleanup of defaults, many kittens died there

PR #36, opened 2023-05-31
Run after install hooks before rebooting

jimmykarily:
otherwise grub is not configured (among other things skipped)

Fixes https://github.com/kairos-io/kairos/issues/1460


internal/agent/install.go
```diff
@@ -300,8 +300,12 @@ func RunInstall(debug bool, options map[string]string) error {
 
 	_, reboot := options["reboot"]
 	_, poweroff := options["poweroff"]
-	installConfig.Reboot = reboot
-	installConfig.PowerOff = poweroff
```
jimmykarily, line 304: Setting these was making the install action to reboot, skipping the AfterInstall hooks that would setup `grubenv` and such.

Itxaka (approved):

PR #38, opened 2023-06-05
:bug: Fix the cleanup of ssh users

Itxaka:
To avoid running the user stuff always in network and try to run it asap


mauromorales (approved):
lgtm

PR #39, opened 2023-06-05
:sparkles: Modify upgrade image flag to accept more formats

Itxaka:
The code is in there to accept several mre types of sources for upgrade, including dir and files.

This patch just exposes that via a new flag called --source and marks the older --image as deprecated, as we can support the same stuff via the new flag.

Also adds the --local flag so we can pull the image from the local repo

### DIFF
--- a/main.go
+++ b/main.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kairos-io/kairos/v2/pkg/utils"
 	"github.com/sirupsen/logrus"
 	"path/filepath"
+	"regexp"
 	"runtime"
 
 	"os"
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -63,9 +64,14 @@ var cmds = []*cli.Command{
 			},
 			&cli.StringFlag{
 				Name:  "image",
-				Usage: "Specify an full image reference, e.g.: quay.io/some/image:tag",
+				Usage: "[DEPRECATED] Specify an full image reference, e.g.: quay.io/some/image:tag",
```
mauromorales, line 67:
```suggestion
				Usage: "[DEPRECATED] Specify a full image reference, e.g.: quay.io/some/image:tag",
```

main.go
```diff
@@ -63,9 +64,14 @@ var cmds = []*cli.Command{
...
+			},
+			&cli.StringFlag{
+				Name:  "source",
+				Usage: "Source for upgrade. Composed of `type:address`. Accepts `file:`,`dir:` or `oci:` for the type of source.\nFor example `file:/var/share/myimage.tar`, `dir:/tmp/extracted` or `oci:repo/image:tag`",
```
mauromorales, line 71: Should we also list `docker:` here?

mauromorales (approved):
very cool

PR #55, opened 2023-06-08
:bug: Read upgrade/install values from config

Itxaka:
We were ignoring the values in the /etc/elemental/config.yaml file that we loaded into viper by not reading those values and their keys into the final spec.

This meant that for example the defautl entry name was being lost as we generated a new install spec from scratch and ignored those values that we read on the config

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/sanity-io/litter"
 	"net/url"
 	"os"
 	"strings"
@@ -101,6 +102,7 @@ func ManualInstall(c string, options map[string]string, strictValidations, debug
 		return err
 	}
 	installConfig.Debug = debug
+	installConfig.Logger.Debugf("Full config: %s\n", litter.Sdump(installConfig))
 
 	return RunInstall(installConfig, options)
 }
@@ -135,6 +137,7 @@ func Install(debug bool, dir ...string) error {
 		return err
 	}
 	installConfig.Debug = debug
+	installConfig.Logger.Debugf("Full config: %s\n", litter.Sdump(installConfig))
 
 	// Reads config, and if present and offline is defined,
 	// runs the installation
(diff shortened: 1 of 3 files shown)


internal/agent/install.go
```diff
@@ -302,7 +305,7 @@ func RunInstall(installConfig *v1.RunConfig, options map[string]string) error {
 	}
 
 	// Generate the installation spec
-	installSpec := elementalConfig.NewInstallSpec(installConfig.Config)
+	installSpec, _ := elementalConfig.ReadInstallSpec(installConfig)
```
Itxaka, line 308: This would read the values loaded into viper and unmarshall those found under the install key.

So a /etc/elemental/config.yaml with the following values:

```
install:
  grub-entry-name: "Kairos"
```

Would be reflected in the InstallSpec properly:

```
DEBU[2023-06-08T09:34:29Z] Full spec: &v1.InstallSpec{
  Target: "/dev/vda",
  Firmware: "bios",
  PartTable: "gpt",
  Partitions: v1.ElementalPartitions{
    BIOS: &v1.Partition{
      Name: "bios",
      FilesystemLabel: "",
      Size: 1,
      FS: "",
      Flags: []string{
        "bios_grub",
      },
      MountPoint: "",
      Path: "",
      Disk: "",
    },
    EFI: nil,
    OEM: &v1.Partition{
      Name: "oem",
      FilesystemLabel: "COS_OEM",
      Size: 64,
      FS: "ext4",
      Flags: []string{}, // p0
      MountPoint: "/run/cos/oem",
      Path: "",
      Disk: "",
    },
    Recovery: &v1.Partition{
      Name: "recovery",
      FilesystemLabel: "COS_RECOVERY",
      Size: 8192,
      FS: "ext4",
      Flags: p0,
      MountPoint: "/run/cos/recovery",
      Path: "",
      Disk: "",
    },
    State: &v1.Partition{
      Name: "state",
      FilesystemLabel: "COS_STATE",
      Size: 15360,
      FS: "ext4",
      Flags: p0,
      MountPoint: "/run/cos/state",
      Path: "",
      Disk: "",
    },
    Persistent: &v1.Partition{
      Name: "persistent",
      FilesystemLabel: "COS_PERSISTENT",
      Size: 0,
      FS: "ext4",
      Flags: p0,
      MountPoint: "/run/cos/persistent",
      Path: "",
      Disk: "",
    },
  },
  ExtraPartitions: nil,
  NoFormat: false,
  Force: false,
  CloudInit: []string{
    "/tmp/xxxx3250348038",
  },
  Iso: "",
  GrubDefEntry: "Kairos",
  Tty: "tty1",
..........
```

mauromorales (approved):
lgtm

PR #56, opened 2023-06-13
:art: Use ext4 by default

Itxaka:
Images were being formated on ext2 for some old reason. Makes no sense to keep them that way, just move everything to ext4 by default


mudler:
the reason is that ext4 has journaling. ext2 doesn't. in systems with RO-only you don't really need journaling, and that means the disk will live longer

mudler:
that being said, I have no other strong preference here, but, would be nice to offer a parameter for that to override the default behavior at this point?

jimmykarily:
If we need some attributes of ext4 but not journaling, it seems it's possible to just disable journaling: https://foxutech.com/how-to-disable-enable-journaling/

Itxaka:
There is also other partitions that we use that are RO that are already on ext4, and we see no issues with that. This is just turning the default in the images to ext4 to have the same fs in all of our partitions. But we can discuss this further on monday meetings

PR #58, opened 2023-06-13
:bug: Behave like a daemon on all scenarios

Itxaka:
When running agent start from a livecd/netboot, the agent sees that it has nothing to do and exits. This behaviour does not sync properly which what is expected from this command, which is to behave a bit like a daemon and stay running in the background.

This patch makes it behave more like a daemon even when we dont need to do anything. This way we can track the agent with supervise-daemon and restart it properly if needed.

This patch uses time.sleep insetead of something cleaner like `select{}` to avoid blocking any other goroutines that migth have started.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -2,9 +2,6 @@ package agent
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
-
 	events "github.com/kairos-io/kairos-sdk/bus"
 	"github.com/kairos-io/kairos-sdk/machine"
 	"github.com/kairos-io/kairos-sdk/utils"
@@ -13,6 +10,10 @@ import (
 	config "github.com/kairos-io/kairos/v2/pkg/config"
 	"github.com/kairos-io/kairos/v2/pkg/config/collector"
 	"github.com/nxadm/tail"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"syscall"
 )
 
 // Run starts the agent provider emitting the bootstrap event.
@@ -35,6 +36,16 @@ func Run(opts ...Option) error {
 	if c.Install != nil && c.Install.Auto && (bf == machine.NetBoot || bf == machine.LiveCDBoot) {
 		// Don't go ahead if we are asked to install from a booting live medium
 		fmt.Println("Agent run aborted. Installation being performed from live medium")
+		// This needs to act like a daemon, so once we reach this point we cannot just exit, we need to stay active
+		// so wait for signals
+		sigc := make(chan os.Signal, 1)
+		signal.Notify(sigc,
+			syscall.SIGHUP,
+			syscall.SIGINT,
+			syscall.SIGTERM,
+			syscall.SIGQUIT)
+		fmt.Println("Listening for signals")
+		<-sigc
 		return nil
 	}
 
@@ -91,5 +102,14 @@ func Run(opts ...Option) error {
 		fmt.Println("Warning: Agent failed, restarting: ", err.Error())
 		return Run(opts...)
 	}
+	// Same here, we got no error, so we stay alive
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc,
+		syscall.SIGHUP,
+		syscall.SIGINT,
+		syscall.SIGTERM,
+		syscall.SIGQUIT)
+	fmt.Println("Listening for signals")
+	<-sigc
 	return err
 }


Itxaka:
Not sure if this is even rigth, but all the scenarios I have seen the agent it behaves like a daemon, so it makes sense to not exit even if on livecd/netboot to avoid supervise-daemon from restarting it again and again.

Itxaka:
weird enough, seems to work as expected ¬_¬

internal/agent/agent.go
```diff
@@ -91,5 +94,7 @@ func Run(opts ...Option) error {
 		fmt.Println("Warning: Agent failed, restarting: ", err.Error())
 		return Run(opts...)
 	}
+	// Same here, we got no error, so we stay alive
+	time.Sleep(time.Duration(1<<63 - 1))
```
jimmykarily, line 98: Maybe it's cleaner if we just listen for "stop" signals. Something like:


```
	sigc := make(chan os.Signal, 1)
	signal.Notify(sigc,
		syscall.SIGHUP,
		syscall.SIGINT,
		syscall.SIGTERM,
		syscall.SIGQUIT)
	fmt.Println("Listening for signals")
	<-sigc
```

Not sure if it makes any difference but it's more explicit and doesn't rely on magic numbers and probably plays nicer with the OS conventions.

jimmykarily: otoh, as soon as the program reaches this point, there is no code path that leads it to do anything useful. In other words, the program keeps running with no possibility to do anything other than exit. If feels weird.

Itxaka: Yes, it is very weird. But we have a path that acts as a daemon and another that doesn't and we can't use the daemon supervisor with both unfortunately.

If we mark it as a one off binary it will stay on starting status forever, if we mark it as a daemon and remove this patch it will keep being restarted and marked as crashed.

No good solution either way 😭

mauromorales: not sure I understand, can you elaborate .... why does it need to stay running doing nothing? just so we can use daemon supervisor? but what good does it make that it stays running if it will not do anything at that point anyway?

Itxaka: well, becuase when we use provider-kairos then it really seems to run as a daemon and it needs to be kept running.

Its just on livecd or on a core kairos where it has nothing to do and just exits and you cant have both ways. If this was systemd it would be easy to make it a "daemon" and then have it exit when doing nothing, but on openrc is much more simpler so it doesnt cover that use case. Either its a one of service (like cos-system-network) or its  a daemon.

Having it behave both ways means that we cant make it work as a proper service. Without this patch, on core kairos it works as expected but on provider it will be shown as "starting" forever, which means that the advantages of supervising it are nil, we will never restart it (although it kind of auto restarts itself).

Im not sure what the proper approach is here to be honest, I just did this patch to test it and discuss to see what we can do if we want to do something about it. 

We can also ignore and live with the fact that in provider-kairos the service will always stay as "starting" and will never be restarted by the supervisor.

Itxaka: oh yeah, another feature of treating it like a proper daemon with supervise is that we can redirect all logging into a file and all errors into a file, which is nice for tests and for end users

mauromorales: gotcha, yeah I guess it's not the only issue we will experience if we handle two different init systems that don't offer the exact same features ... not for now but in the future I wonder what will be more beneficial, to kind of design these processes with openRC in mind first, instead of systemd first ... or just patch this kind of scenarios to mimik a deamon (or some other feature not available)

jimmykarily: being the same binary, doesn't mean it has to always be triggered by the same service definition, right? I mean, we can have 2 different openrc services that call the same binary (kairos-agent) in 2 different ways. That way, we don't expect it to behave like a daemon when it's not supposed to be one. I'm only speaking theoretically though, I haven't seen what the service looks like currently.

Itxaka: > being the same binary, doesn't mean it has to always be triggered by the same service definition, right? I mean, we can have 2 different openrc services that call the same binary (kairos-agent) in 2 different ways. That way, we don't expect it to behave like a daemon when it's not supposed to be one. I'm only speaking theoretically though, I haven't seen what the service looks like currently.

I think it does have to be the same service as it hooks into the events thingies where it behaves differently if there is a provider in the system or not and we don't know beforehand.

It's very confusing to me, I don't really know why it's like this.

Maybe @mudler can share some thoughts on this?

PR #60, opened 2023-06-14
List LVM partitions properly to allow detection of recovery on arm

jimmykarily:
One way to fix: https://github.com/kairos-io/kairos/issues/1486 using `lsblk` instead of `ghw` to list partitions.

Alternatively we could fix ghw to work with lvm (see issue for more).


jimmykarily:
This is a draft. If we like this solution (even temporarily until we upstream a ghw fix), we should move it to the kairos-sdk as a reusable library. This way it can be used here as well: https://github.com/kairos-io/kairos-sdk/pull/28/files#diff-ad95c17c31b539a3e385a23d86bbecf4d8a925ba6679aec2e66b5202ba4c4f81R167

jimmykarily:
I'm not sure we should spend more time moving this to a generic implementation before we try to fix it on ghw. If we decide that an upstream `ghw` solution isn't possible, maybe we re-think about it. I opened this one for review.

jimmykarily:
tests need fixing. back to "in progress".

pkg/cloudinit/layout_test.go
```diff
@@ -251,6 +266,9 @@ BYT;
 				Layout: layout,
 			}
 			runner.SideEffect = func(command string, args ...string) ([]byte, error) {
+				if command == "lsblk" && args[0] == "--list" {
```
mauromorales, line 269: not a hard request but IMO it would be nice to make all the repeated control flows functions, so in the future they only need to get changed in one place

pkg/elementalConfig/config_test.go
```diff
@@ -212,9 +214,12 @@ var _ = Describe("Types", Label("types", "config"), func() {
 						switch cmd {
 						case "cat":
 							return []byte(constants.SystemLabel), nil
-						default:
-							return []byte{}, nil
+						case "lsblk":
```
mauromorales, line 217: even the full switch feels duplicated a bit

mauromorales (approved):
some comments but not blocking ones, LGTM

pkg/elementalConfig/config_test.go
jimmykarily: This is true, but these switch statement are expected shell commands. They may look the same now but at some point, some Go command might introduce a shell command that will make these mock different. Then we will need to switch back to separate mocking per test. I'm not sure it's worth DRYing. After all, while making changes to a test, we should rather not worry that we may break the mocks for other tests.

PR #62, opened 2023-06-19
Install improvements

Itxaka:
- Dont panic if there is no cc.Install
 - Let the config methods be the ones to print the full config on debug
 - Use vfs utils to create tempfiles

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -5,14 +5,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/sanity-io/litter"
 	"net/url"
 	"os"
 	"strings"
 	"syscall"
 	"time"
-
-	"github.com/sirupsen/logrus"
 
 	events "github.com/kairos-io/kairos-sdk/bus"
 	"github.com/kairos-io/kairos-sdk/collector"
@@ -29,6 +26,7 @@ import (
 	qr "github.com/mudler/go-nodepair/qrcode"
 	"github.com/mudler/go-pluggable"
 	"github.com/pterm/pterm"
+	"github.com/spf13/viper"
 	"gopkg.in/yaml.v2"
 )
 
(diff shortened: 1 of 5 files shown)


internal/agent/install.go
```diff
@@ -89,20 +87,24 @@ func ManualInstall(c string, options map[string]string, strictValidations, debug
 	mergeOption(configStr, options)
 
 	if options["device"] == "" {
-		if cc.Install.Device == "" {
-			options["device"] = detectDevice()
-		} else {
-			options["device"] = cc.Install.Device
+		// IF cc is nil we will panic, check beforehand.
+		// If we set no device then we will exit later down the line anyway
+		if cc.Install != nil {
```
jimmykarily, line 92: Merge the 2 `if`s  with `&&` ?

internal/agent/install.go
```diff
@@ -89,20 +87,24 @@ func ManualInstall(c string, options map[string]string, strictValidations, debug
...
+			if cc.Install.Device == "" {
+				options["device"] = detectDevice()
+			} else {
+				options["device"] = cc.Install.Device
+			}
 		}
 	}
 
+	// Set debug from here already, so it's loaded by the ReadConfigRun
+	viper.Set("debug", debug)
```
jimmykarily, line 102: If we are ok with sharing the value of `debug` using viper (like we do for `ReadConfigRun`), why don't we set this at the top level (in `main.go`, manual-install action) so we don't need to pass it as an argument to `ManualInstall`?

internal/agent/install.go
```diff
@@ -131,13 +133,13 @@ func Install(debug bool, dir ...string) error {
 
 	ensureDataSourceReady()
 
+	// Set debug from here already, so it's loaded by the ReadConfigRun
+	viper.Set("debug", debug)
```
jimmykarily, line 137: same

jimmykarily (approved):

Itxaka:
Fuuuu

This had automerge on 😭

I will implement @jimmykarily feedback into a new PR

PR #63, opened 2023-06-19
Drop --local flag for image sources

Itxaka:
The extractor will now auto try to get it from the local daemon first and then drop to remote one like the docker client

Requires: https://github.com/kairos-io/kairos-sdk/pull/32


mauromorales (approved):
lgtm

PR #68, opened 2023-06-22
1541-fix-kairos-agent-lsblk

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/1541

### DIFF
--- a/internal/agent/upgrade.go
+++ b/internal/agent/upgrade.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+
 	"github.com/Masterminds/semver/v3"
 	events "github.com/kairos-io/kairos-sdk/bus"
 	"github.com/kairos-io/kairos-sdk/collector"
@@ -128,7 +129,11 @@ func Upgrade(
 	upgradeConfig.Logger.Debugf("Full config: %s\n", litter.Sdump(upgradeConfig))
 
 	// Generate the upgrade spec
-	upgradeSpec, _ := elementalConfig.ReadUpgradeSpec(upgradeConfig)
+	upgradeSpec, err := elementalConfig.ReadUpgradeSpec(upgradeConfig)
+	if err != nil {
+		return fmt.Errorf("generating the upgrade spec: %w", err)
+	}
+
 	// Add the image source
 	imgSource, err := v1.NewSrcFromURI(img)
 	if err != nil {


internal/agent/upgrade.go
```diff
@@ -128,7 +129,11 @@ func Upgrade(
 	upgradeConfig.Logger.Debugf("Full config: %s\n", litter.Sdump(upgradeConfig))
 
 	// Generate the upgrade spec
-	upgradeSpec, _ := elementalConfig.ReadUpgradeSpec(upgradeConfig)
+	upgradeSpec, err := elementalConfig.ReadUpgradeSpec(upgradeConfig)
```
Itxaka, line 132: FYI even if the underlying issue is fixed, this cant be set like this as it currently is.

See the `ReadUpgradeSpec`  calls the `spec.Sanitize()` which is a validator to make sure that the required values in the spec are loaded.

When used with elemental-cli this was ok, as our cli and config files loaded all needed components, via flags and the elemental configs.
Now that we are hooking up in this from the agent, the agent does not provider all the needed values, so this method will ALWAYS return an error.

In this case, its the `spec.Active.source` values which are always empty, thus the code below on line 142. We load the source that its coming from the agent different configs and then calling `Sanitize` ourselves and catching the error.

Just in case you fix the lsblk problem and keep this in, you will get the same error.

Maybe we should just do like we do with the debug value and set the viper value on the cli directly so we can remove all this parsing from here.

Itxaka: but at this point im wondering if we should have left the elemental-cli cmds in place and just integrated the agent stuff into it lmao

jimmykarily: If sanitize is destined to fail, we shouldn't call it at all (?). After all we call it later again.

Itxaka: but it shouldnt, thats the thing. Currently what we have is a mixmatch of reset/config getting everything from viper and config files and a different cli passing flags around. The idea would be to connect both of those, so flags of the cli get propagated to viper so they are ingested when reading the config. Kind of what we do now with the debug flag, we set that from the cli flag to viper and forget aboout it and the config just reads it.

jimmykarily:
Closing this in favor of either this: https://github.com/kairos-io/kairos-agent/pull/70 or this: https://github.com/kairos-io/kairos-agent/pull/69

PR #69, opened 2023-06-22
Fix lsblk command asterik not being expanded

Itxaka:


pkg/utils/getpartitions.go
```diff
@@ -53,13 +53,45 @@ func GetAllPartitions(runner v1.Runner) (v1.PartitionList, error) {
...
+func PartitionFromLsblk(lsblkOutput string) *v1.Partition {
+	var part = &v1.Partition{}
+	lsblk := &Lsblk{}
+	err := json.Unmarshal([]byte(lsblkOutput), lsblk)
+	if err != nil {
+		return part
+	}
+	if len(lsblk.BlockDevices) == 1 {
+		b := lsblk.BlockDevices[0]
+		part.Name = b.Name
```
Itxaka, line 85: Im missing the parent stuff, is that important? I could not understand why its needed

jimmykarily: There is some place in the code where we use the value of `Disk` or one of the partitions in order to do the "reset" iirc. I need to look it up. Since I didn't know in how many places we rely on the `Disk` to be set correctly (either now or in the future), I tried to populate it with the correct value.

Itxaka: makes sense, added!

jimmykarily:
Closing in favor of: https://github.com/kairos-io/kairos-agent/pull/70

PR #74, opened 2023-07-04
Fix deprecation issues on goreleaser.yaml

mauromorales:
see https://github.com/kairos-io/kairos-agent/actions/runs/5453351229/jobs/9922129053

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -16,14 +16,14 @@ source:
   name_template: '{{ .ProjectName }}-{{ .Tag }}-source'
 archives:
   # Default template uses underscores instead of -
-  - name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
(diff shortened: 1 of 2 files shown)


.goreleaser.yaml
```diff
@@ -16,14 +16,14 @@ source:
...
   # Default template uses underscores instead of -
-  - name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}{{ if .Arm }}v{{ .Arm }}{{ end }}"
-    # TODO: this is deprecated -> https://goreleaser.com/deprecations#archivesreplacements
-    replacements:
-      darwin: Darwin
-      linux: Linux
-      windows: Windows
-      386: i386
-      amd64: x86_64
+  - name_template: >-
```
Itxaka, line 19: well, it may be deprecated but it looks nicer the old way to me lol

Itxaka (approved):
noice

.goreleaser.yaml
mauromorales: indeed, it was much better before IMO

PR #75, opened 2023-07-04
Copy docs into proper place

Itxaka:
They all should be under local dir as there we store all the docs, main page+docs+blog

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -43,7 +43,7 @@ build-kairos-agent:
     RUN apk add upx
     COPY . .
     COPY +webui-deps/node_modules ./internal/webui/public/node_modules
-    COPY github.com/kairos-io/kairos-docs:main+docs/public ./internal/webui/public
+    COPY github.com/kairos-io/kairos-docs:main+docs/public ./internal/webui/public/local
     COPY +version/VERSION ./
     COPY +version/COMMIT ./
     ARG VERSION=$(cat VERSION)


Itxaka:
do not merge before https://github.com/kairos-io/kairos-agent/pull/76 can confirm that we fail on webui properly :D

Earthfile
```diff
@@ -43,7 +43,7 @@ build-kairos-agent:
     RUN apk add upx
     COPY . .
     COPY +webui-deps/node_modules ./internal/webui/public/node_modules
-    COPY github.com/kairos-io/kairos-docs:main+docs/public ./internal/webui/public
+    COPY github.com/kairos-io/kairos-docs:main+docs/public ./internal/webui/public/local
```
mauromorales, line 46: I was having issues with this path and that's why I did the change in https://github.com/kairos-io/kairos-agent/commit/2b13623656e0a257f00d08c82353ef5eba1195eb

Itxaka: interesting, what was the issue? We cannot copy the whole thing under public as that overrides the index with the webpage index as the docs build the docs, the blog and the webpage.

We could copy the docs folder only but then we also need the js folder and css folder probably :D

Itxaka: the index being override is the webui installer btw, so if you build and run the webui and access the localhost:8080 you get the main kairos page.

mauromorales: good question, I don't remember the exact issue ... but I think it breaks the page assets

mauromorales: @Itxaka yeah if I remember correctly the problem is that the website on prod has a certain configuration so when deploying locally it expects the same path

Itxaka: I mean the website on prod is deployed under the kairos-docs repo so it should not affect that?

mauromorales: I think it's related to the full path of the assets ... the website expects them in one place which is defined in config.toml or the default on hugo, not sure ... but then when put under local, there are assets that don't match that full path and show as 404s ... I think if you start the webui and try to look at the docs from there you should see it breaking

Itxaka: The only thing broken is the main page of the docs and their links to architecture, the rest work as expected:
[Screencast from 2023-07-04 17-39-05.webm](https://github.com/kairos-io/kairos-agent/assets/1447686/94c62bc3-b65f-49cb-aecc-76348708c37b)

And Im not sure that is not our issue, i.e. every other link is relative while those appear different...

mauromorales: ok cool thanks for checking, maybe it's not a biggie then

mauromorales (approved):

Earthfile
Itxaka: mmmh, gonna have a look at this further to see if it can be fixed in the docs to refer to everything relative to the current dir

Itxaka: seems that this can be fixed by creating the links with `{{< ref "../upgrade/kubernetes" >}}` which also has the benefit of making hugo check the link to see if its valid :D

That should fix this issue and make it more portable!

mauromorales: very cool 😎

PR #76, opened 2023-07-04
Fail on webui failures

Itxaka:


Itxaka:
Failed as expected, nice

PR #78, opened 2023-07-07
Search for lvm partitions on upgrade

Itxaka:
This was fixed on reset but not for upgrade.

When dealing with lvm partitions (arm rpi for example) some partitions migth be on lvm so we cant find them and we panic

This adds the same extra checks for lvm partitions to the upgrade spec

Fixes https://github.com/kairos-io/kairos/issues/1593


mauromorales (approved):
lgtm

PR #82, opened 2023-07-12
Load elemental config/spec from cloud config

Itxaka:
First round.

Picks the config from the cloud config IF its there, otherwise defaults to proper values that we used to ship with the elemental config.

This allows to set any of the old elemental config options via cloud config directly as such:

```
#cloud-config
debug: true
install:
  reboot: true
  grub-entry-name: "MyCustomOS"
  system:
    size: 5000
  recovery-system:
    size: 1400000

cloud-init-paths:
- /run/initramfs/cos-state
```

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -96,11 +96,7 @@ func ManualInstall(c string, options map[string]string, strictValidations bool) 
 			}
 		}
 	}
-	installConfig, err := elementalConfig.ReadConfigRunFromCloudConfig(configStr)
-	if err != nil {
-		return err
-	}
-	return RunInstall(installConfig, options)
+	return RunInstall(options)
 }
 
 func Install(dir ...string) error {
@@ -139,13 +135,7 @@ func Install(dir ...string) error {
 		r["device"] = cc.Install.Device
 		mergeOption(configStr, r)
 
-		// Load the installation Config from the system
-		installConfig, err := elementalConfig.ReadConfigRunFromCloudConfig(configStr)
-		if err != nil {
-			return err
-		}
-
-		err = RunInstall(installConfig, r)
+		err = RunInstall(r)
 		if err != nil {
 			return err
 		}
@@ -237,15 +227,9 @@ func Install(dir ...string) error {
 
 	r["cc"] = config.AddHeader(header, string(out))
 
-	// Load the installation Config from the system
-	installConfig, err := elementalConfig.ReadConfigRunFromCloudConfig(r["cc"])
-	if err != nil {
-		return err
-	}
-
 	pterm.Info.Println("Starting installation")
 
-	if err := RunInstall(installConfig, r); err != nil {
+	if err := RunInstall(r); err != nil {
 		return err
 	}
 
@@ -262,14 +246,7 @@ func Install(dir ...string) error {
 	return nil
 }
 
-func RunInstall(installConfig *v1.RunConfig, options map[string]string) error {
-	f, err := elementalUtils.TempFile(installConfig.Fs, "", "kairos-install-config-xxx.yaml")
-	if err != nil {
-		installConfig.Logger.Error("Error creating temporal file for install config: %s\n", err.Error())
-		return err
-	}
-	defer os.RemoveAll(f.Name())
-
+func RunInstall(options map[string]string) error {
 	cloudInit, ok := options["cc"]
 	if !ok {
 		fmt.Println("cloudInit must be specified among options")
@@ -288,6 +265,19 @@ func RunInstall(installConfig *v1.RunConfig, options map[string]string) error {
 	// Does it make sense anymore? We can now expose the whole options of elemental directly
 	env := append(c.Install.Env, c.Env...)
 	utils.SetEnv(env)
+
+	// Load the installation Config from the system
+	installConfig, err := elementalConfig.ReadConfigRunFromCloudConfig(cloudInit)
+	if err != nil {
+		return err
+	}
+
+	f, err := elementalUtils.TempFile(installConfig.Fs, "", "kairos-install-config-xxx.yaml")
+	if err != nil {
+		installConfig.Logger.Error("Error creating temporal file for install config: %s\n", err.Error())
+		return err
+	}
+	defer os.RemoveAll(f.Name())
 
 	err = os.WriteFile(f.Name(), []byte(cloudInit), os.ModePerm)
 	if err != nil {
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -11,10 +11,6 @@ import (
 	"syscall"
 	"time"
 
-	events "github.com/kairos-io/kairos-sdk/bus"
-	"github.com/kairos-io/kairos-sdk/collector"
-	"github.com/kairos-io/kairos-sdk/machine"
-	"github.com/kairos-io/kairos-sdk/utils"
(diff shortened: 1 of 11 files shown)


pkg/constants/constants.go
```diff
@@ -24,7 +24,7 @@ import (
 const (
 	GrubConf                = "/etc/cos/grub.cfg"
 	GrubOEMEnv              = "grub_oem_env"
-	GrubDefEntry            = "cOS"
+	GrubDefEntry            = "Kairos"
```
Itxaka, line 27: About time....

pkg/elementalConfig/config.go
```diff
@@ -222,18 +222,21 @@ func NewInstallSpec(cfg v1.Config) *v1.InstallSpec {
 		recoveryImg.Source = v1.NewFileSrc(recoveryImgFile)
 		recoveryImg.FS = constants.SquashFs
 		recoveryImg.File = filepath.Join(constants.RecoveryDir, "cOS", constants.RecoverySquashFile)
+		recoveryImg.Size = constants.ImgSize
```
Itxaka, line 225: This were defaults that were missing, hence the necessity of adding them in our shipped config.yaml for elemental under kairos.....

pkg/types/v1/config.go
```diff
@@ -122,12 +122,13 @@ func (c *Config) Sanitize() error {
...
-	PowerOff       bool     `yaml:"poweroff,omitempty" mapstructure:"poweroff"`
-	CloudInitPaths []string `yaml:"cloud-init-paths,omitempty" mapstructure:"cloud-init-paths"`
-	EjectCD        bool     `yaml:"eject-cd,omitempty" mapstructure:"eject-cd"`
+	Debug           bool     `yaml:"debug,omitempty" mapstructure:"debug"`
+	Strict          bool     `yaml:"strict,omitempty" mapstructure:"strict"`
+	Reboot          bool     `yaml:"reboot,omitempty" mapstructure:"reboot"`
+	PowerOff        bool     `yaml:"poweroff,omitempty" mapstructure:"poweroff"`
+	CloudInitPaths  []string `yaml:"cloud-init-paths,omitempty" mapstructure:"cloud-init-paths"`
+	EjectCD         bool     `yaml:"eject-cd,omitempty" mapstructure:"eject-cd"`
+	FullCloudConfig string   // Stores the full cloud config used to generate the spec afterwards
```
Itxaka, line 131: This is for ease of use. We can reuse the cloudconfig embedded here to generate the spec afterwards without having to pass it again. And maybe it will be helpful down the line to have it there?

Itxaka:
Do not merge until the next point release has been done.

main.go
```diff
@@ -582,7 +590,15 @@ The validate command expects a configuration file as its only argument. Local fi
...
-			cfg, err := elementalConfig.ReadConfigRun("/etc/elemental")
+			config, err := config.Scan(collector.Directories(configScanDir...), collector.NoLogs)
+			if err != nil {
+				return err
+			}
+			cc, err := config.String()
+			if err != nil {
+				return err
+			}
+			cfg, err := elementalConfig.ReadConfigRunFromCloudConfig(cc)
```
mauromorales, line 601: but the pattern seems to be similar so yeah I think this method, or a similar one should be able to consume config directly

internal/agent/install.go
```diff
@@ -286,6 +266,19 @@ func RunInstall(installConfig *v1.RunConfig, options map[string]string) error {
 	env := append(c.Install.Env, c.Env...)
 	utils.SetEnv(env)
 
+	// Load the installation Config from the system
```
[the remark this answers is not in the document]
Itxaka: what part? The file creation is only done here and the config stuff is already a function.

mauromorales:
did the review from Golang, seems like it duplicates a few things :p

main.go
```diff
@@ -535,7 +535,15 @@ The validate command expects a configuration file as its only argument. Local fi
...
```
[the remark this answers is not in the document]
Itxaka: we could pass indeed the config directly and do the extract inside but in sooo many place we have already the cloud-init string ready that it seemed duplicating the extraction...not sure what to do here to be honest

pkg/elementalConfig/config.go
Itxaka: yep yep, lots of duplicated code around, needs some refactoring

pkg/elementalConfig/config.go
```diff
@@ -539,62 +544,31 @@ func NewBuildConfig(opts ...GenericOptions) *v1.BuildConfig {
...
+	// Store the full cloud-config in here so we can reuse it afterwards
+	cfg.FullCloudConfig = cc
 	err = cfg.Sanitize()
 	cfg.Logger.Debugf("Full config loaded: %s", litter.Sdump(cfg))
 	return cfg, err
 }
 
-func ReadInstallSpec(r *v1.RunConfig) (*v1.InstallSpec, error) {
+// ReadInstallSpecFromCloudConfig generates an installation spec from the raw cloud-config data
+func ReadInstallSpecFromCloudConfig(r *v1.RunConfig) (*v1.InstallSpec, error) {
```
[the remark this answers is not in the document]
Itxaka: To be honest, this should be extracted into its own function and receive a runconfig + the key so we dont have duplication. 

Could also merge the config+spec into one function and return both, so get the cloud-config+action (upgrade,install,reset) string and return the config and the spec directly instead of doing it in 2 different places

pkg/types/v1/config.go
```diff
@@ -122,12 +122,13 @@ func (c *Config) Sanitize() error {
 }
 
 type RunConfig struct {
-	Debug          bool     `yaml:"strict,omitempty" mapstructure:"debug"`
-	Strict         bool     `yaml:"strict,omitempty" mapstructure:"strict"`
-	Reboot         bool     `yaml:"reboot,omitempty" mapstructure:"reboot"`
-	PowerOff       bool     `yaml:"poweroff,omitempty" mapstructure:"poweroff"`
-	CloudInitPaths []string `yaml:"cloud-init-paths,omitempty" mapstructure:"cloud-init-paths"`
-	EjectCD        bool     `yaml:"eject-cd,omitempty" mapstructure:"eject-cd"`
+	Debug           bool     `yaml:"debug,omitempty" mapstructure:"debug"`
```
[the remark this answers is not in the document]
Itxaka: I did not catch this, it just started failing LOL

main.go
Itxaka: yeah, this needs another refactor afterwards to merge them more and more and consolidate stuff.

pkg/types/v1/config.go
Itxaka: yes, I think so but could not make it work without a LOT of changes. Next step is merging more things together once we got 1 source of config and start de-duplicating a lot of things.

This all needs a lot of care.

mauromorales (approved):

Itxaka:
@mauromorales updated with less duplication and consolidating the config and spec get to a single function so we have to change it only on one place. Also accepts now the AgentConfig directly

mauromorales (approved):
lgtm

pkg/elementalConfig/config.go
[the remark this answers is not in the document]
Itxaka: ah yes, this was done already, marking it as resolved!

PR #83, opened 2023-07-12
Create .codecov.yaml

Itxaka:
Seems like the old config was still stored on codecov, but once we changed the module name it reset and now we got codecov commenting on PRs and failing the PRs due to not reaching the required code coverage ¬_¬


mauromorales (approved):
I was wondering what that was about, good catch

PR #85, opened 2023-07-17
Fix oem mountpoint not available on upgrade

Itxaka:
This causes the chroot stages to not be complete as they could not bind mount the current system /oem into the chroot due not being able to infer the oem partition mountpoint, so any cloud-init files in there would be ignored.

Also removes the log downgrade for yip that was useful under yip 0.11.x. Now the logging is nicer and if you want debug logging you get it.

### DIFF
--- a/pkg/elementalConfig/config.go
+++ b/pkg/elementalConfig/config.go
@@ -369,6 +369,11 @@ func NewUpgradeSpec(cfg v1.Config) (*v1.UpgradeSpec, error) {
 		}
 	}
 
+	// If we have oem in the system, but it has no mountpoint
+	if ep.OEM != nil && ep.OEM.MountPoint == "" {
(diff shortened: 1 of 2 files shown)


pkg/elementalConfig/config.go
```diff
@@ -369,6 +369,14 @@ func NewUpgradeSpec(cfg v1.Config) (*v1.UpgradeSpec, error) {
 		}
 	}
 
+	// If we have oem in the system
+	if ep.OEM != nil {
+		// But it has no mountpoint
+		if ep.OEM.MountPoint == "" {
```
mauromorales, line 375: since both come from `ep.OEM`, it would be nice to have a predicate function for this e.g. `ep.MissingMountPoint()`

Itxaka: done!

mauromorales (approved):
lgtm

PR #90, opened 2023-07-20
Fix releases coming from the bus

Itxaka:
When we get the releases from github we sort them adn reverse them to get the proper latest version the first on the list.

But when we receive the list from the bus manager we have no idea if they are coming sorted or not.

This patch makes sure we sort and reverse the list to get the latest in the first position

Fixes https://github.com/kairos-io/kairos/issues/1611

### DIFF
--- a/internal/agent/upgrade.go
+++ b/internal/agent/upgrade.go
@@ -4,17 +4,18 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
 
 	"github.com/Masterminds/semver/v3"
-	events "github.com/kairos-io/kairos-sdk/bus"
-	"github.com/kairos-io/kairos-sdk/collector"
-	"github.com/kairos-io/kairos-sdk/utils"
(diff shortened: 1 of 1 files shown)


Itxaka:
```
localhost:~$ ./kairos-agent upgrade
Searching for releases
Latest release is v2.3.0-k3sv1.27.3-k3s1
Are you sure you want to upgrade to this release? (y/n) n
localhost:~$ 
```

This shows the necessity of making this more complex:
 - should we offer the upgrade to the same k3s version that the current system?
 - should the kairos version have priority over the k3s version for the upgrade?
 - if there is not a clear path for upgrade (1.23 in current system but it was dropped for the next release) should we offer the upgrade to the next k8 version (1.24) or to the latest?

Also this means that the provider need to manage the versions more nicely and have a matrix of kairos version - k8s version instead of just 1 single version

mauromorales (approved):
lgtm

PR #91, opened 2023-07-20
Merge data from events into reset

Itxaka:
Otherwise we have 2 different sources for the cloud-config options and we have no idea what to do with them.

This patch extract the 3 different config options that we are able to
override and sets them before running the reset

Should fix https://github.com/kairos-io/kairos/issues/1626

cc @santhoshdaivajna this would change the way you send the data a bit.

Instead of sending the option and an empty value for it you need to explicitly set the value of the option to either true or false.

Any empty values would be ignored.
Any other keys would be ignored.
Any values that differ from true/false would be ignored.
Default values for the reset spec should be good, we should not need to override anything.

so instead of sending:

```
response, _ := json.Marshal(map[string]string{
		"strict":           "",
		"reset-persistent": "",
		"system.uri":       "dir:/",
	})
```

you would send:

```
response, _ := json.Marshal(map[string]string{
		"strict":           "false",
		"reset-persistent": "true",
		"reset-oem":       "false",
	})
```

Notice that we only allow 3 options to be override:

 - reset-persistent -> True by default, would format the persistent partition
 - reset-oem -> False by default, would format the oem partition
 - strict -> False by default, would fail during cloud-config stages (before-reset, after-reset-chroot, after-reset) if any issues are encountered

We believe that the default values since 2.0.0 are in a sane state, so you can also drop and not send anything. Persistent should be formatted during a reset so you get a clena node, oem should be kept as it has configuration files for the system and strict should be disabled as there can always be minor issues in the default cloud-config file shipped with the system (there is actually a couple of issues already which we have no fix for for now)

### DIFF
--- a/pkg/elementalConfig/config.go
+++ b/pkg/elementalConfig/config.go
@@ -495,12 +495,13 @@ func NewResetSpec(cfg v1.Config) (*v1.ResetSpec, error) {
 
 	activeFile := filepath.Join(ep.State.MountPoint, "cOS", constants.ActiveImgFile)
 	return &v1.ResetSpec{
-		Target:       target,
-		Partitions:   ep,
-		Efi:          efiExists,
-		GrubDefEntry: constants.GrubDefEntry,
-		GrubConf:     constants.GrubConf,
-		Tty:          constants.DefaultTty,
+		Target:           target,
+		Partitions:       ep,
+		Efi:              efiExists,
+		GrubDefEntry:     constants.GrubDefEntry,
+		GrubConf:         constants.GrubConf,
+		Tty:              constants.DefaultTty,
+		FormatPersistent: true,
 		Active: v1.Image{
 			Label:      constants.ActiveLabel,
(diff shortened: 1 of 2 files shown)


pkg/elementalConfig/config.go
```diff
@@ -495,12 +495,13 @@ func NewResetSpec(cfg v1.Config) (*v1.ResetSpec, error) {
...
-		GrubDefEntry: constants.GrubDefEntry,
-		GrubConf:     constants.GrubConf,
-		Tty:          constants.DefaultTty,
+		Target:           target,
+		Partitions:       ep,
+		Efi:              efiExists,
+		GrubDefEntry:     constants.GrubDefEntry,
+		GrubConf:         constants.GrubConf,
+		Tty:              constants.DefaultTty,
+		FormatPersistent: true,
```
Itxaka, line 504: This makes persistent to be formatted by default. As it should be

santhoshdaivajna (approved):

PR #96, opened 2023-07-21
Rework reboot/shutdown to use the hooks

Itxaka:
We were manually calling the reboot/shutdown from different places with no organization nor checks.

This patch makes it so we use the hooks for reset/upgrade to reboot/poweroff.

This allows us first, to configure the individual install/reset/upgrade actions with their own reboot/shutdown config, so tehy can behave differently, and to have a generic way of doing so.

This changes the hooks signatures to pass a spec so we can inspect it for values for the given action.

This also means that we can expand the hooks for example to set the grub entry name and inspect the actual spec to get that info.


Itxaka:
@mudler do we have something else using the hooks? I dont think so no?

internal/agent/hooks/lifecycle.go
```diff
@@ -1,18 +1,22 @@
...
-	if c.Install.Reboot {
+func (s Lifecycle) Run(c config.Config, spec v1.Spec) error {
+	if spec.ShouldReboot() {
+		time.Sleep(5)
 		utils.Reboot()
 	}
 
-	if c.Install.Poweroff {
+	if spec.ShouldShutdown() {
+		time.Sleep(5)
```
mauromorales, line 19: should we add a message, saying the action will take place in 5 seconds?

mauromorales (approved):
nice change!

internal/agent/hooks/lifecycle.go
Itxaka: I would but this would just print to stdout and I dont like that at all, because if you want to log to a file or make the install quiet then you get different outputs to different places  :/

I hope to soon(ish) merge the agent config+elemental config into just 1 main config and once that its done, because on the hook we have access to the config object we will be able to access teh config.Logger and log to it respecting the output, level, etc....

mauromorales: yup agree, better wait for logger, maybe some TODO then? but up to you

PR #98, opened 2023-07-21
Cleanup and simplify elementalConfig

Itxaka:

### DIFF
--- a/pkg/elementalConfig/config.go
+++ b/pkg/elementalConfig/config.go
@@ -19,9 +19,6 @@ package elementalConfig
 import (
 	"fmt"
 	"gopkg.in/yaml.v3"
-	"io"
-	"io/fs"
-	"os"
 	"path/filepath"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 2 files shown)


pkg/elementalConfig/config.go
```diff
@@ -553,20 +525,24 @@ func NewBuildConfig(opts ...GenericOptions) *v1.BuildConfig {
...
 		return nil, err
 	}
 
-	configLogger(cfg.Logger, cfg.Fs)
 	err = yaml.Unmarshal([]byte(cc), &cfg)
 	if err != nil {
 		return nil, err
 	}
-	// Store the full cloud-config in here so we can reuse it afterwards
+	// If we got debug enabled via cloud config, set it on viper so its available everywhere
```
mauromorales, line 540: nice!!

pkg/elementalConfig/config.go
```diff
@@ -42,83 +39,62 @@ import (
...
-func WithArch(arch string) func(r *v1.Config) error {
-	return func(r *v1.Config) error {
-		r.Arch = arch
-		return nil
-	}
-}
-
-func WithPlatform(platform string) func(r *v1.Config) error {
-	return func(r *v1.Config) error {
+func WithPlatform(platform string) func(r *v1.Config) {
```
mauromorales, line 86: Not sure I understand why we can get rid of the error

pkg/elementalConfig/config_test.go
```diff
@@ -92,6 +95,21 @@ var _ = Describe("Types", Label("types", "config"), func() {
 				Expect(c.Mounter).To(Equal(mounter))
 				Expect(c.Runner).ToNot(BeNil())
 			})
+			It("defaults to sane platform is the platform is broken", func() {
```
mauromorales, line 98:
```suggestion
			It("defaults to sane platform if the platform is broken", func() {
```

pkg/elementalConfig/config.go
Itxaka: Mainly because we dont use it anywhere in a proper way so its useless. This functions are just for setting private values on the struct and that should never fail.

mauromorales (approved):
lgtm

PR #100, opened 2023-07-24
Rework install a bit more

Itxaka:
Also fixes some of the loading pars from cloud config into elemental specs

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -62,7 +62,7 @@ func mergeOption(cloudConfig string, r map[string]string) {
 	}
 }
 
-func ManualInstall(c string, options map[string]string, strictValidations bool) error {
+func ManualInstall(c, device string, reboot, poweroff, strictValidations bool) error {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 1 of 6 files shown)


internal/agent/install.go
```diff
@@ -288,7 +236,12 @@ func RunInstall(options map[string]string) error {
 	}
 	defer os.RemoveAll(f.Name())
 
-	err = os.WriteFile(f.Name(), []byte(cloudInit), os.ModePerm)
+	ccstring, err := c.String()
+	if err != nil {
+		installConfig.Logger.Error("Error creating temporal file for install config: %s\n", err.Error())
```
jimmykarily, line 241:
```suggestion
		installConfig.Logger.Error("Error creating temporary file for install config: %s\n", err.Error())
```

internal/agent/install.go
```diff
@@ -127,22 +120,18 @@ func Install(dir ...string) error {
 	// runs the installation
```
jimmykarily, line 120: I'm glad to see all this weird juggling of options in this function to be gone now. I just wonder if they were addressing some very specific use case or they are just relics from when we had multiple configs around. I hope it's just the latter.

pkg/types/v1/config.go
```diff
@@ -146,7 +146,7 @@ func (r *RunConfig) Sanitize() error {
 
 // InstallSpec struct represents all the installation action details
 type InstallSpec struct {
-	Target          string              `yaml:"target,omitempty" mapstructure:"target"`
+	Target          string              `yaml:"device,omitempty" mapstructure:"device"`
```
jimmykarily, line 149: Is this a "breaking" change? Is it something we should warn the users about? (e.g. if they have configs that used `target`).

jimmykarily (approved):
Nice cleanup. Let's see what the end 2 end tests say about it on the kairos side.

pkg/types/v1/config.go
Itxaka: not really, this is the target in the elemental config, which now was unused by agent as it had its own target.

What this does it to sync it with the agent so it can load it from the cloud config directly. This is ONLY used on the grub install IIRC and we set it manually with the data from the agent Config already. This just allows us to have it in sync automatically :D

PR #101, opened 2023-07-24
Drop runconfig in favour of config

Itxaka:
As we no longer need different configs for both build stuff and run
    stuff we can merge back the runconfig into config, which will make it
    easier to merge it into the agentconfig

### DIFF
--- a/internal/agent/hooks/runstage.go
+++ b/internal/agent/hooks/runstage.go
@@ -16,7 +16,7 @@ func (r RunStage) Run(c config.Config, _ v1.Spec) error {
 	if err != nil {
 		cfg.Logger.Errorf("Error reading config: %s\n", err)
 	}
-	_ = utils.RunStage(&cfg.Config, "kairos-install.after", cfg.Strict, cfg.CloudInitPaths...)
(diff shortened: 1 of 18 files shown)


Itxaka:
Based of https://github.com/kairos-io/kairos-agent/pull/100 so it needs that to be merged.

Itxaka:
#100 merged and this rebased!

internal/agent/hooks/runstage.go
```diff
@@ -16,7 +16,7 @@ func (r RunStage) Run(c config.Config, _ v1.Spec) error {
 	if err != nil {
 		cfg.Logger.Errorf("Error reading config: %s\n", err)
 	}
-	_ = utils.RunStage(&cfg.Config, "kairos-install.after", cfg.Strict, cfg.CloudInitPaths...)
+	_ = utils.RunStage(cfg, "kairos-install.after")
```
Itxaka, line 19: as we now pass the full cfg in here, we no longer need to add the strict or cloudinitpaths as a separate entity, we can just use the cfg object ot get those :D

mauromorales (approved):
lgtm

PR #102, opened 2023-07-24
Merge elemental config into agent config

Itxaka:
- Just one place for everything
 - Dropped the extra module for elemental config
 - Some shared utils moved into their own modules to avoid circular
   dependencies
 - logger now available everywhere!
 - Config has the whole thing on it, less mangling of the config
 - spec now in the config module

builds upon #100 and #101 so those need to go in before this one


internal/agent/hooks/hook.go
```diff
@@ -10,7 +10,6 @@ type Interface interface {
 }
 
 var AfterInstall = []Interface{
-	&RunStage{},    // Shells out to stages defined from the container image
```
Itxaka, line 13: This was removed and run directly on the action like the others.

internal/agent/install.go
```diff
@@ -76,6 +66,8 @@ func ManualInstall(c, device string, reboot, poweroff, strictValidations bool) e
 		return err
 	}
 
+	fmt.Print(litter.Sdump(cc))
```
mauromorales, line 69: is this intended?

main.go
```diff
@@ -527,11 +526,11 @@ The validate command expects a configuration file as its only argument. Local fi
 		},
 		Action: func(c *cli.Context) error {
 			stage := c.Args().First()
-			config, err := config.Scan(collector.Directories(configScanDir...), collector.NoLogs)
+			config, err := agentConfig.Scan(collector.Directories(configScanDir...), collector.NoLogs)
 			if err != nil {
 				return err
 			}
-			cfg, err := elementalConfig.ReadConfigRunFromAgentConfig(config)
+			cfg, err := agentConfig.ReadConfigRunFromAgentConfig(config)
```
mauromorales, line 533: this is returning a `v1.Config`, why do we still need that distinction vs `config` before? ... and if we have to keep it maybe it helps to name them something different to make this obvious

pkg/config/config.go
```diff
@@ -36,6 +42,59 @@ type Install struct {
 	BindMounts             []string          `yaml:"bind_mounts,omitempty"`
 }
 
+func NewConfig(opts ...GenericOptions) *Config {
+	log := v1.NewLogger()
+
+	defaultPlatform, err := v1.NewPlatformFromArch(runtime.GOARCH)
```
mauromorales, line 48: Can we name this something other than default? maybe underlyingPlatform or hostPlatform, or something like that?

pkg/config/config.go
```diff
@@ -46,6 +105,149 @@ type Config struct {
...
+		return err
+	}
+
+	return nil
+}
+
+// LoadInstallState loads the state.yaml file and unmarshals it to an InstallState object
+func (c Config) LoadInstallState() (*v1.InstallState, error) {
+	installState := &v1.InstallState{}
+	data, err := c.Fs.ReadFile(filepath.Join(constants.RunningStateDir, constants.InstallStateFile))
```
mauromorales, line 156: I see this was copied from before so no need to change in this PR but it feels like the full function could be a method of config

pkg/utils/grub.go
```diff
@@ -18,21 +18,22 @@ package utils
 
 import (
 	"fmt"
+	agentConfig "github.com/kairos-io/kairos-agent/v2/pkg/config"
+	fs2 "github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
```
mauromorales, line 22: maybe fsutils would fit better than fs2

Itxaka: Good catch, wrong rename when moving the package

internal/agent/install.go
Itxaka: Nope! Good catch

main.go
Itxaka: umm, impossible, v1.Config does not exist anymore, there is now just one Config struct in the whole repo. Where did you see that?

pkg/config/config.go
Itxaka: done!

pkg/config/config.go
Itxaka: ? isnt it already a method of Config?

main.go
mauromorales: if I understand correctly 

```
config, err := agentConfig.Scan(collector.Directories(configScanDir...), collector.NoLogs)
```

see https://github.com/kairos-io/kairos-agent/pull/102/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R529 

returns a `config.Config`

while `cfg, err := agentConfig.ReadConfigRunFromAgentConfig(config)`

returns a types `v1.Config` which I saw was already replaced for some of the other cases

pkg/config/config.go
mauromorales: but not a big deal this one

main.go
Itxaka: ohh I see. Both are returning a config. There is no longer a v1.Config, but this is being called twice for no reason probably!

Itxaka: and this returns the same config in both cases:

`func ReadConfigRunFromAgentConfig(c *Config) (*Config, error) `

Itxaka: in any case, sometimes the config needs to be initialized in a different config to fullfill some of the required stuff.

for example the `agentConfig.Scan` will return a config, yes, but that config wont have the logger on it for example.

I tihnk the fix here is to hijack the Scan to create the config with the `Newconfig` method, so we dont have an empty config.Logger for example.

Itxaka: good catch here, seems like I was doing this twice due to initializing the Config object on the scan directly. Now it does so all the config defaults are set from scratch and much more code can be removed, yay!

Itxaka:
This is now broken lmao

main.go
Itxaka: ok, fixed and even improved!

mauromorales: cool cool cool

mauromorales (approved):
lgtm

PR #103, opened 2023-07-25
Fix config loading, install device and call sanitize

Itxaka:
Syncs the cfg.install.device to the installspec.target As on manual-install the device can actually come from a flag and its not on the cloud-config, we need to initialize the InstallSpec with the value of the cfg.Install.Device as by that time we already have the final value of the install target.

Also calls sanitize on config and specs and warns if anything goes wrong


Itxaka:
Directly merging as main is broken

PR #104, opened 2023-07-26
Add tests for Platform

Itxaka:

### DIFF
--- a/pkg/types/v1/common_test.go
+++ b/pkg/types/v1/common_test.go
@@ -17,9 +17,12 @@ limitations under the License.
 package v1_test
 
 import (
+	"fmt"
+	"github.com/kairos-io/kairos-agent/v2/pkg/constants"
 	v1 "github.com/kairos-io/kairos-agent/v2/pkg/types/v1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
 )
 
 var _ = Describe("Types", Label("types", "common"), func() {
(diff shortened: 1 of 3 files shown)


pkg/types/v1/common_test.go
```diff
@@ -112,5 +115,95 @@ var _ = Describe("Types", Label("types", "common"), func() {
...
+			Expect(platform.OS).To(Equal("linux"))
+			Expect(platform.Arch).To(Equal(constants.ArchArm64))
+			Expect(platform.GolangArch).To(Equal(constants.ArchArm64))
+			platform, err = v1.NewPlatform("linux", constants.Archx86)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(platform.OS).To(Equal("linux"))
+			Expect(platform.Arch).To(Equal(constants.Archx86))
+			Expect(platform.GolangArch).To(Equal(constants.ArchAmd64))
+		})
+		It("Does not check the validity of the os", func() {
```
mauromorales, line 131: we currently can only run on linux, no?

Itxaka: yes, but its not even checked. Not sure what to do with this, but it may require more validation in the future. For now its just testing whats in there.

Itxaka: I kind of remember this being useful to build cross-platform isos/raw images or pull images from other platforms, so it could be used to pull a risc image from a registry for example. So that is why there is no validation and its a free-form thing I think

pkg/types/v1/config_test.go
```diff
@@ -224,4 +225,225 @@ var _ = Describe("Types", Label("types", "config"), func() {
...
+					Expect(err.Error()).To(ContainSubstring("undefined recovery partition"))
+				})
+				It("passes if state and source are ready", func() {
+					spec.Recovery.Source = v1.NewFileSrc("/tmp")
+					spec.Partitions.Recovery = &v1.Partition{
+						MountPoint: "/tmp",
+					}
+					err := spec.Sanitize()
+					Expect(err).ToNot(HaveOccurred())
+				})
```
mauromorales, line 444: These seem so similar between the diferent `Describe` cases that mabye they could be written once in the beforeEach, and then the differences changed in a context in the justBeforeEach

mauromorales (approved):
lgtm

pkg/types/v1/config_test.go
Itxaka: Indeed butmore importantly to me it shows that we have too many things in common in the sanitize methods that could be extracted into a single function and reused or a generic sanitize created for the common parts!

PR #105, opened 2023-07-26
Drop logrusWrapper

Itxaka:
The only reason this was there is to have compatibility with luet due to installing from channel sources.

Now that there is no channel sources, there is no point in having it anymore as its not used.

### DIFF
--- a/pkg/types/v1/logger.go
+++ b/pkg/types/v1/logger.go
@@ -18,12 +18,8 @@ package v1
 
 import (
 	"bytes"
-	"fmt"
-	"io"
-	"regexp"
-	"strings"
-
 	log "github.com/sirupsen/logrus"
+	"io"
 )
 
 // Logger is the interface we want for our logger, so we can plug different ones easily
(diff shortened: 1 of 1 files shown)


pkg/types/v1/logger.go
```diff
@@ -33,7 +29,6 @@ type Logger interface {
 	Debug(...interface{})
 	Error(...interface{})
 	Fatal(...interface{})
-	Success(...interface{})
```
mauromorales, line 36: Why does the interface need to change? shouldn't whatever other tool we use just follow the interface?

Itxaka: The other tool that used this was luet and changing that was much more difficult than adapting our logger to confirm to let standards. Otherwise we got the default let logger which logged very differently from 3lemental and produced a very ugly output between both apps.

Now that we don't use luet in here we can safely drop our logger interface to the minimum needed as we are the only ones using it. 

Those methods were only used by luet, anything under elemental was using the simple debug,info,fatal,warn facilities

Itxaka: Mainly to avoid the dissonance between both loggers, we adapted ours and passed it to luet so we could manage it's output properly, but let needed this interfaces methods.

mauromorales: gotcha

mauromorales (approved):
lgtm

PR #106, opened 2023-08-01
Rework install workflow for event stuff

Itxaka:
As we modified the source of config for the install to be a single object, we left out the event driven configuration

This caused the config to be alwasy empty when triggered from an event (like registration with qr code)

This patch makes it so the config is a declared var so we can override that var with the data coming from the events easily and the RunInstall will get a final fully Config from all the sources

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -6,8 +6,10 @@ import (
 	"errors"
 	"fmt"
 	fsutils "github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
+	"github.com/sanity-io/litter"
 	"net/url"
 	"os"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/agent/install.go
```diff
@@ -82,6 +84,10 @@ func ManualInstall(c, device string, reboot, poweroff, strictValidations bool) e
 }
 
 func Install(dir ...string) error {
+	var cc *config.Config
+	var err error
+
+	bus.Manager.Initialize()
```
Itxaka, line 90: this is already done on Main during command init, but I seen that reset and upgrade have it also in their functions, in case they are called outside of the main function somehow?

Seems like adding it doesnt hurt and could prevent issues.

Itxaka:
This fixes the provider tests

mauromorales (approved):
lgtm

PR #108, opened 2023-08-02
Extra cleanup for install partitions

Itxaka:
As we now allow overriding from cloud-config, users could go over and overide the partitions labels, and we need those labels to be fixed, otherwise we wont be able to identify any partitions at all.

This override the partition label and filesystem label to their default values on the installSpec sanitize method


pkg/types/v1/config.go
```diff
@@ -97,6 +97,18 @@ func (i *InstallSpec) Sanitize() error {
...
+	// In case we got values override from the config files, we need to make sure to set the labels properly
+	// Otherwise we might break the system as we rely on know labels for a lot of things.
+	i.Partitions.OEM.FilesystemLabel = constants.OEMLabel
+	i.Partitions.OEM.Name = constants.OEMPartName
+	i.Partitions.State.FilesystemLabel = constants.StateLabel
+	i.Partitions.State.Name = constants.StatePartName
+	i.Partitions.Persistent.FilesystemLabel = constants.PersistentLabel
+	i.Partitions.Persistent.Name = constants.PersistentPartName
+	i.Partitions.Recovery.FilesystemLabel = constants.RecoveryLabel
+	i.Partitions.Recovery.Name = constants.RecoveryPartName
```
mauromorales, line 110: Can we make this a method of partitions? would be nice to call it something like:

```
i.Partitions.SetDefaultLabels()
```

Feel free to choose a better name, but I think it reduces the size of the outer scope function and keeps the details on how to set the default inside partitions

pkg/types/v1/config.go
```diff
@@ -97,6 +97,18 @@ func (i *InstallSpec) Sanitize() error {
 	if extraPartsSizeCheck == 1 && i.Partitions.Persistent.Size == 0 {
 		return fmt.Errorf("both persistent partition and extra partitions have size set to 0. Only one partition can have its size set to 0 which means that it will take all the available disk space in the device")
 	}
+
+	// In case we got values override from the config files, we need to make sure to set the labels properly
+	// Otherwise we might break the system as we rely on know labels for a lot of things.
```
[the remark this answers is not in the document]
Itxaka: I meant `known labels` maybe `fixed labels` is more clear?

pkg/types/v1/config.go
Itxaka: Good point. It started with one line and then it kept growing with more :D

mauromorales (approved):
LGTM

PR #109, opened 2023-08-02
Allow creating dirs in rootfs via cc

Itxaka:
Currently we have an issue that during install,reset,upgrade the rootfs is not able to be managed properly to create extra dirs.

This can be an issue for apps that write to non standard paths and once installed you have to rely in either building your custom image with those dirs or playing with the cc stages to create the dirs.

This new facility allows an user to configure paths to be created during install, upgrade and reset as we have full access to the RW rootfs.

### DIFF
--- a/pkg/action/common.go
+++ b/pkg/action/common.go
@@ -18,7 +18,10 @@ package action
 
 import (
 	config "github.com/kairos-io/kairos-agent/v2/pkg/config"
+	cnst "github.com/kairos-io/kairos-agent/v2/pkg/constants"
 	"github.com/kairos-io/kairos-agent/v2/pkg/utils"
+	fsutils "github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
+	"path/filepath"
 )
 
 // Hook is RunStage wrapper that only adds logic to ignore errors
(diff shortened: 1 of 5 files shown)


pkg/action/upgrade.go
```diff
@@ -181,6 +181,17 @@ func (u *UpgradeAction) Run() (err error) {
...
+	// Create extra dirs in rootfs as afterwards this will be impossible due to RO system
+	for _, d := range u.spec.ExtraDirsRootfs {
+		if exists, _ := fsutils.Exists(u.config.Fs, filepath.Join(upgradeImg.MountPoint, d)); !exists {
+			u.config.Logger.Debugf("Creating extra dir %s under %s", d, upgradeImg.MountPoint)
+			err = u.config.Fs.Mkdir(filepath.Join(upgradeImg.MountPoint, d), constants.DirPerm)
+			if err != nil {
+				u.config.Logger.Warnf("Failure creating extra dir %s in rootfs at %s", d, upgradeImg.MountPoint)
+			}
+		}
+	}
```
mauromorales, line 193: These three are the same no? can we make it a single function, so we don't have to keep them up to date whenever one changes?

Itxaka: Yep, indeed. I got a bit confused due to the upgrade source being different, but in the end it's the same thing, a mountpoint and mkdir.


Will extract it. Although I'm wary of extracting too much stuff into the utils as that causes circular deps...

pkg/action/common.go
```diff
@@ -39,3 +42,15 @@ func ChrootHook(config *config.Config, hook string, chrootDir string, bindMounts
 	}
 	return utils.ChrootedCallback(config, chrootDir, bindMounts, callback)
 }
+
+func createExtraDirsInRootfs(cfg *config.Config, extradirs []string, target string) {
+	for _, d := range extradirs {
+		if exists, _ := fsutils.Exists(cfg.Fs, filepath.Join(target, d)); !exists {
+			cfg.Logger.Debugf("Creating extra dir %s under %s", d, target)
+			err := cfg.Fs.Mkdir(filepath.Join(target, d), cnst.DirPerm)
```
jimmykarily, line 50: Given the Fs is passed through the config, it might be easy to mock it and create a test for this new functionality.

Sorry for the late review, feel free to skip or do it in another PR :)

Itxaka: umm good point, will do!

PR #110, opened 2023-08-03
Make unattended reset work

Itxaka:
Our current example for cloud images on the docs on how to use the reset with a cloud-config and a stage does not actually work as the tty adquisition and such seems to be failing due tto not being properly run in a tty.

This pathc introduces 2 new flags to the reset command.

reboot: Enable reboot after reset, overriding any config in the system. This is only a enable switch, so if the flag is disabled but the config says reset it will NOT override it.

unattended: Disables printing stuff into the screen, locking the terminal and falling back to a new tty on run. Also enables fast mode automatically.


main.go
```diff
@@ -446,9 +446,21 @@ This command is meant to be used from the boot GRUB menu, but can likely be used
...
+				Usage: "Enable automated reboot after reset. Has precedence over any config in the system.",
+			},
+			&cli.BoolFlag{
+				Name:  "unattended",
+				Usage: "Do not wait for user input and provide ttys after reset. Also sets the fast mode (do not wait 60 seconds before reset)",
+			},
+		},
 		Action: func(c *cli.Context) error {
+			reboot := c.Bool("reboot")
+			unattended := c.Bool("unattended")
```
mauromorales, line 461: will this ever be part of the config? or is it always going to be flags? just wondering if we should do a similar thing of merging it to the existing config

Itxaka: flags is a way of overriding the config values in the system, so they are explicitly passed. As this is a very specific flag for a very specific scenario, Im not sure if we should merge it back into the config.

Basically for me this means, stop behaving as usual, do this special thing with overrides. So not sure. Maybe its helpful to have it merged there as a standard value in the config to avoid the tty stuff? 

But still the problem here for me is that there is no connection between flags and config. Like with viper+cobra is that you can unmarshall the flags into a config and do it in order so your flags dont require manual handling at all, you just have a method to do the binding thing but here the flags have no link to the config and it needs manual stuff :(

Itxaka:
This works as expected for cloud images. Should not affect anything else.

main.go
mauromorales: yeah indeed, we should really consider that I think

Itxaka: I think we either be able to get teh flags to propagate to the config somehow or move to use cobra+viper for full integration

mauromorales (approved):
LGTM

PR #113, opened 2023-08-03
Make interactive installer use the collector as well

Itxaka:
Its the only method not using it so it was missing some nice things on it

Also this makes ALL methods now use the collector so we can fix stuff on all of them at the same time


mauromorales (approved):
lgtm

PR #122, opened 2023-08-07
Auto calculate image size on actions

Itxaka:
Isntead of using the default static values, during the spec creation we can try to infer the actual size of image, dir or OCI image directly and fall back to the defaults if we are not able to do so.

This is compatible with users overriding the image sizes via cloud-config as its done much earlier, while creating the initial spec, even before loading the values from cloud config into the spec.

This requires https://github.com/kairos-io/kairos-sdk/pull/39 + a new sdk version

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -18,8 +18,7 @@ package config
 
 import (
 	"fmt"
-	"github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
-	"github.com/kairos-io/kairos-agent/v2/pkg/utils/partitions"
+	"io/fs"
 	"path/filepath"
 	"reflect"
 	"strings"
@@ -27,6 +26,8 @@ import (
 	"github.com/kairos-io/kairos-agent/v2/internal/common"
 	"github.com/kairos-io/kairos-agent/v2/pkg/constants"
 	v1 "github.com/kairos-io/kairos-agent/v2/pkg/types/v1"
+	"github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
+	"github.com/kairos-io/kairos-agent/v2/pkg/utils/partitions"
 	"github.com/mitchellh/mapstructure"
 	"github.com/sanity-io/litter"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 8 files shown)


pkg/utils/fs/fs.go
```diff
@@ -198,7 +199,7 @@ func WalkDirFs(fs v1.FS, root string, fn fs.WalkDirFunc) error {
 	} else {
 		err = walkDir(fs, root, &statDirEntry{info}, fn)
 	}
-	if err == filepath.SkipDir {
+	if errors.Is(err, filepath.SkipDir) {
```
Itxaka, line 202: this is not connected but my IDE was complaining.

mauromorales (approved):
lgtm

pkg/config/spec.go
```diff
@@ -400,6 +438,44 @@ func ReadInstallSpecFromConfig(c *Config) (*v1.InstallSpec, error) {
...
+
+	case source.IsFile():
+		file, err := config.Fs.Stat(source.Value())
+		if err == nil {
+			size = file.Size()
+		}
+	}
+	// Normalize size to Mb before returning and add 100Mb to round the size from bytes to mb+extra files like grub stuff
+	if size != 0 {
+		size = (size / 1000 / 1000) + 100
```
jimmykarily, line 474: This is integer division and it could be "throwing away" more than you expect:

```
	i := 8999999
	fmt.Println(i / 1000 / 1000)
```
In the example above, the return value is just `8`, while `9` is surely a value closer to truth. Just adding 100Mb when we have thrown away `999999` won't cut it :). Btw, what's the unit of `size`? It looks like it's Terabytes since we divide twice by 1000 to get Mb. Is that right? Maybe name the variable something like `sizeTB` to make it clear.

jimmykarily: Sorry, it's still morning :man_facepalming: , `size` is bytes. Ignore this. Worst case scenario, the division throws away 1 Mb and then we add 100 on top. All good.

Itxaka: It's Mb.

And I had so many issues with this. I tried to make them float to not lost stuff but o was getting a zero size at the end. I'm not sure what I was doing wrong here, all sizes from the size check come in bytes but our partitioner works with Mb so we need to translate it here.

I would try again today to improve this, see if I'm more smart today than yesterday 😂

pkg/config/spec.go
```diff
@@ -237,13 +250,26 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
...
+	}
+
+	// Get the actual source size to calculate the image size and partitions size
+	size, err := GetSourceSize(cfg, spec.Active.Source)
+	if err != nil {
+		cfg.Logger.Warnf("Failed to infer size for images: %s", err.Error())
+	} else {
+		cfg.Logger.Infof("Setting image size to %dMb", size)
+		spec.Active.Size = uint(size)
+		spec.Passive.Size = uint(size)
```
jimmykarily, line 268: I may be missing something here. When we upgrade, the passive image is the old one right? Could it be that the old one and the new one have different sizes? Is it ok to infer the size of the passive from the source?

Itxaka: Passive is overwritten with the current active, so mainly a rename of active.img to passive.img and this value is ignored and only the active one is used.

So yeah, this is actually ignored and not used. Maybe I should infer the active image size and use that for passive? Before this, the default size was shown, so it's mainly esthetic, to show the values in the dump of the spec.

I will change it so it gets the default constant as it should not be override by this, it should have the default values, even if not used.

PR #124, opened 2023-08-08
Restore loading env from config file

Itxaka:
Useful to pass and set env vars during install for things like the docker config location


Itxaka:
This was probably lost during the converge of the cloud config into config and such.

internal/agent/install.go
```diff
@@ -226,6 +226,9 @@ func Install(dir ...string) error {
 }
 
 func RunInstall(c *config.Config) error {
+	utils.SetEnv(c.Env)
+	utils.SetEnv(c.Install.Env)
```
mauromorales, line 230: do these add on top of the other?

mauromorales: Gotcha, thanks! And yeah I guess long term having per action might be a good idea just like you have flags per action

mauromorales (approved):
LGTM

PR #125, opened 2023-08-08
Fix goreleaser artifact names

Itxaka:
Otherwise the archive files have weird punctuation:` kairos-agent-.v2.2.3-Linux-x86_64.tar.gz`


mauromorales (approved):
glad you found it!

PR #129, opened 2023-08-17
Several improvements for size calculation

Itxaka:
- Pass the image source and recoverySource values via viper so we can get them in the upgrade spec directly
 - Calculate the image size for docker image properly as its uncompressed
 - Add 1Gb to the state partition so future images can grow
 - Umount persistent partition
 - Set recovery size only if we are upgrading recovery


pkg/config/spec.go
```diff
@@ -258,15 +260,31 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
 		State:      installState,
 	}
 
-	// Get the actual source size to calculate the image size and partitions size
+	source := viper.GetString("upgradeSource")
+	recoveryUpgrade := viper.GetBool("upgradeRecovery")
+	if source != "" {
+		imgSource, err := v1.NewSrcFromURI(source)
+		if err == nil {
+			if recoveryUpgrade {
```
mauromorales, line 268: not for now but just an idea to architect this a bit differently, IMO it would be nice to have a strategy pattern instead of doing the `if recoveryUpgrade` here and there

mauromorales (approved):
LGTM

pkg/config/spec.go
Itxaka: well, its  switch becuase we need to pass the source into the proper place, mainly so it shows only in the proper place.

I could bindly add the upgrade source into both but then it would appear in the final spec, which I dont really like as it could lead to unknown issues.

upgrade source for recovery should only be set if we are upgrading recovery otherwise it should be an empty source and the sanitize method would fail for the proper reason.

PR #130, opened 2023-08-18
Adjust size to take into account upgrades

Itxaka:
Upgrades of active use the state partition to create the transition.img which will then substitute the active.img
Recovery upgrades use the recovery partition for the same.

We were not taking into account that then we need more space in there for the image to reside there while doing the upgrade which could lead to space exhaustion easily, so this calculates the partitions sizes taking into account that we may neeed double the space of the default image. Still keeps the extra 1Gb in there so we can grow the images in the future.


mauromorales (approved):
lgtm

Itxaka:
stupid size test

PR #131, opened 2023-08-18
Fix manual-install reboot/poweroff flags

Itxaka:
We were only checking if the spec has the reboot/poweroff enabled but on manual-install we set the values via flags into the Config.Install struct so we need to check that as well

Fixes: https://github.com/kairos-io/kairos/issues/1739


mauromorales (approved):
lgtm

PR #132, opened 2023-08-21
Fix recovery image calculation

Itxaka:
We were basica the calculation on the size of the Active.Image but on recovery, the calculation should be done of the Recovery.Image.

That resulted in a value of 0 for the size AND no error.

Instead move the calculation above when we have access the the sourceImage directly and we can calculate it in one go.


mauromorales (approved):
lgtm

PR #133, opened 2023-09-01
Track active/passive boot for upgrade workflow

Itxaka:
Track what are we booting from during upgrade in order to not backup active into passive if we are booting from passive.

If we are booting from active or recovery the workflow keeps as it was, backup active into passive, and then move the transition image into active.

Fixes: https://github.com/kairos-io/kairos/issues/1286
Needs: https://github.com/kairos-io/kairos-sdk/pull/46

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -18,15 +18,16 @@ package action
 
 import (
 	"fmt"
-	agentConfig "github.com/kairos-io/kairos-agent/v2/pkg/config"
-	"github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
 	"path/filepath"
 	"time"
 
+	agentConfig "github.com/kairos-io/kairos-agent/v2/pkg/config"
 	"github.com/kairos-io/kairos-agent/v2/pkg/constants"
 	"github.com/kairos-io/kairos-agent/v2/pkg/elemental"
 	v1 "github.com/kairos-io/kairos-agent/v2/pkg/types/v1"
 	"github.com/kairos-io/kairos-agent/v2/pkg/utils"
+	"github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
+	"github.com/kairos-io/kairos-sdk/state"
 )
 
 // UpgradeAction represents the struct that will run the upgrade from start to finish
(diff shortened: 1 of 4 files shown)


pkg/utils/fs/fs.go
```diff
@@ -248,3 +248,32 @@ func readDir(fs v1.FS, dirname string) ([]fs.DirEntry, error) {
...
+const (
+	Active   = "active_boot"
+	Passive  = "passive_boot"
+	Recovery = "recovery_boot"
+	LiveCD   = "livecd_boot"
+	Unknown  = "unknown"
+)
+
+// DetectBoot Here until we move this to sdk to accept a fs.
+func DetectBoot(fs v1.FS) (string, error) {
```
Itxaka, line 261: This is currently on the sdk but the version in there does not accept a vfs so it always looks on the real system for the /proc/cmdline which makes it impossible to test properly

Itxaka:
if https://github.com/kairos-io/kairos-sdk/pull/46 is merged before this, then I can drop the detectboot from here and use the one in the sdk directly

pkg/action/upgrade.go
```diff
@@ -230,9 +241,10 @@ func (u *UpgradeAction) Run() (err error) {
 		return err
 	}
 
-	// If not upgrading recovery, backup active into passive
-	if !u.spec.RecoveryUpgrade {
-		//TODO this step could be part of elemental package
+	// If not upgrading recovery and booting from non passive, backup active into passive
+	// We dont want to overwrite passive if we are booting from passive as it could mean that active is broken and we would
+	// be overriding a working passive with a broken/unknown  active
+	if u.spec.RecoveryUpgrade == false && bootedFrom != state.Passive {
```
jimmykarily, line 247: Is there a way to test this?

jimmykarily: I mean in the test suite (not manually)

Itxaka: in the end2end or in the unit tests? Unit tests its easy, will add once I see this passing with the new sdk!

end2end may be more complicated as its a whole new test when booting from passive..

jimmykarily (approved):

pkg/action/upgrade.go
Itxaka: Fixed the test. As there was no command mock for copying the active into passive, the command was doing nothing, so on master it directly tried to copy it.


Now if the upgrade tries to copy active into passive the test will directly panic

Itxaka: tested the test on master which doesnt have the newer workflow and it panics as expected. With this patch it never tries to override the  passive.

PR #144, opened 2023-09-12
Also install fonts when installing grub

Itxaka:
Should fix the second part of https://github.com/kairos-io/kairos/issues/1767

### DIFF
--- a/pkg/utils/grub.go
+++ b/pkg/utils/grub.go
@@ -134,7 +134,7 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool, 
 		// as they were before. We now use the bundled grub.efi provided by the shim package
 		g.config.Logger.Infof("Generating grub files for efi on %s", target)
 		var foundModules bool
-		for _, m := range []string{"loopback.mod", "squash4.mod", "xzio.mod", "gzio.mod"} {
+		for _, m := range constants.GetGrubModules() {
 			err = fsutils.WalkDirFs(g.config.Fs, rootDir, func(path string, d fs.DirEntry, err error) error {
 				if err != nil {
 					return err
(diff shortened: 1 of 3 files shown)


pkg/utils/grub.go
```diff
@@ -160,6 +160,34 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
...
+					err = g.config.Fs.WriteFile(fileWriteName, fileContent, cnst.FilePerm)
+					if err != nil {
+						return fmt.Errorf("error writing %s: %s", fileWriteName, err)
+					}
+					foundFonts = true
+					return nil
+				}
+				return err
+			})
+			if !foundFonts {
```
jimmykarily, line 185: Shouldn't this be outside the for loop block?

pkg/utils/grub.go
```diff
@@ -160,6 +160,34 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
 			}
 		}
 
+		var foundFonts bool
```
jimmykarily, line 163: The whole thing can be extracted to an `installGrubFonts` function (or some other fitting name).

pkg/utils/grub.go
Itxaka: not really, I just missed adding the name here. We only want to warn the user about it if we cant find  font X so it needs to be under the loop.

pkg/utils/grub.go
Itxaka: just used once in here, not reusable. I dont think it needs the poor visibility of extracting into a function :)

Itxaka:
test failures have nothing to do with this patch, its the http test that its returning a 403

pkg/utils/grub.go
jimmykarily: Extracting code to a function has other benefits (testing separately, making sure it's self-contained and that no variables are touched by this code when they shouldn't, etc). Self-contained tasks should live in self-contained functions but we don't seem to follow this rule anywhere else in the code. This is not the proper place to introduce new rules. I'm fine leaving it as is.

pkg/utils/grub.go
Itxaka: I just copied the code from the function above when we do the same for the grub modules. I cant really remember now why its done like this, but pretty sure it had its purpose....which I cant remember now XD

Itxaka: Trying to remember here, maybe it was done that way because we needed all modules and any of those failing should fail, but in this case the fonts are not really needed, its just nice to haver, so maybe the approach has to be different. No need for this variable anymore, can just warn inside the loop and call it a day....

pkg/utils/grub.go
```diff
@@ -225,3 +227,36 @@ func (g Grub) SetPersistentVariables(grubEnvFile string, vars map[string]string)
 	}
 	return nil
 }
+
+// copyGrubFonts will try to finds and copy the needed grub fonts into the system
+// rootdir is the dir where to search for the fonts
+// bootdir is the base dir where they will be copied
+func copyGrubFonts(cfg *agentConfig.Config, rootDir, bootDir, systemgrub string) {
```
Itxaka, line 234: cc @jimmykarily :)

jimmykarily (approved):

PR #145, opened 2023-09-14
Check root permissions in commands that need it

jimmykarily:
Otherwise the error shown is not helpful (it fails too late with confusing errors).


jimmykarily:
pull-image already had a check: https://github.com/kairos-io/kairos-agent/blob/274701a3aca3bc495b1549ad3a173f56e02f7e3d/main.go#L576-L578

The function I added is in a different package and not exported and I preferred not to re-use it. The alternative is to export it and use it in main.go too.

Itxaka (changes requested):
I would prefer this to be done on the main go inside the cmds with the `Before:` stanza, as otherwise a lot of work is done on the commands sometimes (like reading and preparing config and specs) for no reason and in fact that work could as well fail due to not enough permissions.

internal/agent/upgrade.go
```diff
@@ -51,6 +52,10 @@ func ListReleases(includePrereleases bool) semver.Collection {
 
 func Upgrade(
 	version, source string, force, strictValidations bool, dirs []string, preReleases, upgradeRecovery bool) error {
+	if err := checkRoot(); err != nil {
```
Itxaka, line 55: leftover

internal/agent/interactive_install.go
```diff
@@ -130,6 +130,10 @@ func detectDevice() string {
 }
 
 func InteractiveInstall(debug, spawnShell bool) error {
+	if err := checkRoot(); err != nil {
```
Itxaka, line 133: leftover

jimmykarily: oops, others too. Fixed.

Itxaka (approved):
noice

PR #150, opened 2023-09-25
Pass the install event to the collector directly

Itxaka:
Instead of saving files and whatnot, we can directly pass the event cc
data as a reader and have it merged into the final config

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
@@ -108,22 +107,10 @@ func Install(dir ...string) error {
 		if err != nil {
 			fmt.Println(err)
 		}
-		// dump data into a dir so the collector can pick it up properly
 		cloudConfig, exists := r["cc"]
 		if exists {
-			tmpdir, err := os.MkdirTemp("", "kairos-install-")
-			if err == nil {
-				err = os.WriteFile(filepath.Join(tmpdir, "kairos-event-install-data.yaml"), []byte(cloudConfig), os.ModePerm)
-				if err != nil {
-					fmt.Printf("could not write event cloud init: %s\n", err.Error())
-				}
-				// Append to default dirs so we read from all sources
-				dir = append(dir, tmpdir)
-				// override cc with our new config object from the scan, so it's updated for the RunInstall function
-				cc, _ = config.Scan(collector.Directories(dir...), collector.MergeBootLine, collector.NoLogs)
-			} else {
-				fmt.Printf("could not create temp dir: %s\n", err.Error())
-			}
+			// Re-read the full config and add the config coming from the event
+			cc, _ = config.Scan(collector.Directories(dir...), collector.Readers(strings.NewReader(cloudConfig)), collector.MergeBootLine, collector.NoLogs)
 		}
 	})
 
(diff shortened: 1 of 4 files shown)


Itxaka:
May fix https://github.com/kairos-io/kairos/issues/1853 but as usual with the events, its a bit complicated to test if indeed its received that properly and parsing it properly......

@mudler any idea how can I manually test this?

Itxaka:
Im guessing witht he provider you can run `kairos-agent notify 'agent.install' 'JSON_PAYLOAD'` ?

internal/agent/install.go
```diff
@@ -113,6 +113,10 @@ func Install(dir ...string) error {
 		if exists {
 			tmpdir, err := os.MkdirTemp("", "kairos-install-")
 			if err == nil {
+				// Add cloud config header if necessary
+				if !strings.Contains(cloudConfig, "#cloud-config") {
+					cloudConfig = fmt.Sprintf("#cloud-config\n%s", cloudConfig)
```
mauromorales, line 118: shouldn't we add the config in the config instead of here? otherwise the check for the config is not really that useful IMO

Itxaka: This is the object that its marshalled into json and pumped via the event:

```
type CloudInit struct {
	Install       kairos.Install            `yaml:"install,omitempty" json:"install,omitempty"`
	Options       map[string]string         `yaml:"options,omitempty" json:"options,omitempty"`
	Cluster       clusterplugin.Cluster     `yaml:"cluster,omitempty" json:"cluster,omitempty"`
	Stages        map[string][]schema.Stage `yaml:"stages,omitempty" json:"stages,omitempty"`
	Stylus        Stylus                    `yaml:"stylus,omitempty" json:"stylus,omitempty"`
	StylusPackage *string                   `yaml:"stylusPackage,omitempty" json:"stylusPackage,omitempty"`
	Env           []string                  `yaml:"env,omitempty" json:"env,omitempty"`
}

```

So ti kinds of make sense that what we receive its not with the header and what we save as cloud config needs the header to be read back up.

I wonder if we could make the collector read dirs and Reader objects directly so we could apss that data directly and skip the header stuff for those files.

Itxaka: Im wondering if we should just try to unmarshall the cloudConfig into an agentconfig.Config object directly and merge it on top of the base config which should be there already....

That would make this simpler and would make the event config end up on top of the others as expected....

jimmykarily: I don't remember how I used the mock provider back then but I assume I just copied that to [the directory where providers are expected](https://github.com/kairos-io/kairos-agent/blob/9f46af255f056c974d335dd305eac837fb012efb/internal/bus/bus.go#L35) and let kairos-agent run it.

internal/agent/install.go
```diff
@@ -113,6 +113,10 @@ func Install(dir ...string) error {
 		if exists {
 			tmpdir, err := os.MkdirTemp("", "kairos-install-")
 			if err == nil {
+				// Add cloud config header if necessary
+				if !strings.Contains(cloudConfig, "#cloud-config") {
```
jimmykarily, line 117: Maybe we should adapt and use this: https://github.com/kairos-io/kairos-agent/blob/9f46af255f056c974d335dd305eac837fb012efb/pkg/config/config.go#L273

Itxaka: This is not a Config object so we cannot access that. This is just a plain string.

Itxaka:
@jimmykarily using the new reader functionality of the collector.

collector commit is not merged into master!

jimmykarily (approved):
Nice!

mauromorales (approved):

Itxaka:
collector patch merged, sdk commit updated

PR #152, opened 2023-09-26
1837 Add `--source` to various commands

jimmykarily:
Fixes: https://github.com/kairos-io/kairos/issues/1837


internal/agent/install.go
```diff
@@ -250,7 +250,28 @@ func RunInstall(c *config.Config) error {
...
 	// Get the source of the installation if we are overriding it
-	if c.Install.Image != "" {
+	if sourceImg != "" {
+		imgSource, err := v1.NewSrcFromURI(sourceImg)
+		if err != nil {
+			return err
+		}
+		installSpec.Active.Source = imgSource
+
+		// TODO: Why only setting active source above? What about size?
```
Itxaka, line 260: on install, only the active is used as a source, as it makes no sense having separated sources for all 3 during an install. There is only 1 source for a clean install and its used for all three images (active,passive,recovery)

hence, no need to set anything else. You can see this on the `NewInstallSpec`

Itxaka:
Currently on upgrade, we use viper to set the var and then recover it on the spec generation.

Is not the best thing ever to be honest, but it helps to set it once and forget about it instead of passing it via all the methods.

viper is a singleton, so once you set a value on it, its available everywhere. This allows us to set it even when parsing the source on the  maing.go and then on the spec generation we can just get it and use it.

Unless we are able to somehow bind the flag to the spec, I thikn we should do just that, set the source just after validating it and then on the spec generation, get it and use it. Less moving pieces all around and a single way of setting it and getting it for all commands.

jimmykarily:
> viper is a singleton, so once you set a value on it, its available everywhere. 

A singleton is not much better than global variables: https://en.wikipedia.org/wiki/Singleton_pattern#Criticism
and there are many reasons to avoid global variables (and singletons). Let me see if I can find a cleaner way to pass it around, otherwise I'll just use it the same way.

(btw, I'm suspecting [1800](https://github.com/kairos-io/kairos/issues/1800)) has something to do with how we pass these through viper but I can't reproduce it yet)

jimmykarily:
@Itxaka now that the collector [accepts readers as well](https://github.com/kairos-io/kairos-agent/pull/150/files#diff-bdceadb6b1ed909e97670e80e5251016ad0826cff05ec4cac9ed83add43cd992R113), maybe a better idea is to create a config with all the command line options and pass it to the collector as a reader. This way to propagates down to everything.

After all, these command line arguments are really configuration and the sooner we "merge" it the better, because all the consumers of the config will have access to it. What do you think?

Itxaka:
> @Itxaka now that the collector [accepts readers as well](https://github.com/kairos-io/kairos-agent/pull/150/files#diff-bdceadb6b1ed909e97670e80e5251016ad0826cff05ec4cac9ed83add43cd992R113), maybe a better idea is to create a config with all the command line options and pass it to the collector as a reader. This way to propagates down to everything.
> 
> After all, these command line arguments are really configuration and the sooner we "merge" it the better, because all the consumers of the config will have access to it. What do you think?

Hell yeah, that is an awesome idea!

internal/agent/upgrade_test.go
```diff
@@ -0,0 +1,47 @@
+package agent
```
jimmykarily, line 1: This test fails on CI because it doesn't stub anything so it tried to mount some partition.

jimmykarily:
The PR in its current state seems to work (manual QA). I want to give another take on the tests. It's currently not easy to test because even creating the UpgradeSpec involves mounting partitions and such.

pkg/types/v1/common.go
```diff
@@ -35,31 +35,31 @@ const (
 
 // ImageSource represents the source from where an image is created for easy identification
 type ImageSource struct {
-	source  string
-	srcType string
+	Source  string
+	SrcType string
```
jimmykarily, line 39: These need to be exported for `yip` to fill them in when unmarshalling

Itxaka: Do they? They don't get unmarshalled via viper directly no? Don't they get filled by the newsrcfromuri method? Maybe we can get away without exporting them?


Not that it matters much to be honest, just trying to understand how it worked before...

jimmykarily: You are right. I was trying to set the fields directly in the yaml initially before I realized we have the custom unmarhaller. Reverted.

internal/agent/upgrade_test.go
jimmykarily: I removed the test. Make this code testable requires a lot more refactoring. Might happen on another PR.

internal/agent/install.go
```diff
@@ -331,3 +339,14 @@ func prepareConfiguration(ctx context.Context, source string) (string, error) {
 
 	return f.Name(), nil
 }
+
+func generateInstallConfForCLIArgs(source string) string {
```
mauromorales, line 343: it's hard to tell for me what kind of sources I can pass here? but down in the generated response it seems it is a URL, it's a minor detail but maybe we can extend it with `sourceURL` or such? then it also helps when writing a caller with the IDE

internal/agent/install.go
```diff
@@ -53,16 +53,21 @@ func displayInfo(agentConfig *Config) {
 	}
 }
 
-func ManualInstall(c, device string, reboot, poweroff, strictValidations bool) error {
+func ManualInstall(c, sourceImg, device string, reboot, poweroff, strictValidations bool) error {
```
mauromorales, line 56: I would also extend it to somethign like `sourceImgURL` to make it clearer to consume

internal/agent/install.go
```diff
@@ -83,7 +88,7 @@ func ManualInstall(c, device string, reboot, poweroff, strictValidations bool) e
 	return RunInstall(cc)
 }
 
-func Install(dir ...string) error {
+func Install(sourceImg string, dir ...string) error {
```
mauromorales, line 91: also here

internal/agent/interactive_install.go
```diff
@@ -129,7 +129,7 @@ func detectDevice() string {
 	return preferedDevice
 }
 
-func InteractiveInstall(debug, spawnShell bool) error {
+func InteractiveInstall(debug, spawnShell bool, sourceImg string) error {
```
mauromorales, line 132: same comment, sorry for the spam

internal/agent/install.go
```diff
@@ -331,3 +339,14 @@ func prepareConfiguration(ctx context.Context, source string) (string, error) {
 
 	return f.Name(), nil
 }
+
+func generateInstallConfForCLIArgs(source string) string {
+	if source == "" {
+		return ""
+	}
+
+	return fmt.Sprintf(`install:
```
mauromorales, line 348: maybe it's worth validating if source is a real uri?

internal/agent/upgrade.go
```diff
@@ -148,17 +95,133 @@ func determineUpgradeImage(version string) (string, error) {
...
-		return "", fmt.Errorf("can't find IMAGE_REPO key under /etc/os-release %w", err)
+		return nil, fmt.Errorf("can't find IMAGE_REPO key under /etc/os-release %w", err)
+	}
+
+	return v1.NewSrcFromURI(fmt.Sprintf("%s:%s", registry, version))
+}
+
+// generateUpgradeConfForCLIArgs creates a kairos configuration for `--source` and `--recovery`
+// command line arguments. It will be added to the rest of the configurations.
+func generateUpgradeConfForCLIArgs(source string, upgradeRecovery bool) (string, error) {
```
mauromorales, line 115: same comment, though I see why source might have been prefered, though I still think in functions it's worth the detail of what that source represents

internal/agent/install.go
mauromorales: I see a validation at the gathering of the source, so this comment can be ignored

jimmykarily: This part of the code makes a config out of `--source` cli argument. If its' invalid, they user will get whatever error they would get if they put it in the Kairos config themselves. As such, this function's responsibility is only to make it part of the configuration and not validate it (which should happen later).

internal/agent/upgrade.go
jimmykarily: Yes "source" was chosen because it matches the command line argument exactly: `--source`. I can rename the functions though.

internal/agent/install.go
jimmykarily: renamed this to `sourceImageURL` like I did for the rest of the functions. At least it's more clear that it is the same value we are passing around.

main.go
```diff
@@ -686,3 +711,19 @@ func checkRoot() error {
 
 	return nil
 }
+
+func validateSource(source string) error {
+	if source == "" {
+		return nil
+	}
+
+	r, err := regexp.Compile(`^oci:|dir:|file:`)
```
mauromorales, line 720: what about docker? https://github.com/kairos-io/kairos-agent/blob/1170576dcce4018a3dffddf500295c3ac3a64a46/pkg/types/v1/common.go#L30

internal/agent/upgrade.go
jimmykarily: renamed to `sourceImageURL`

internal/agent/install.go
jimmykarily: done

internal/agent/install.go
jimmykarily: done

main.go
mauromorales: and somehow it feels like these two checks should be related instead of having to set at each file

main.go
```diff
@@ -436,10 +449,22 @@ See also https://kairos.io/docs/installation/qrcode/ for documentation.
...
+			if err := validateSource(c.String("source")); err != nil {
+				return err
+			}
+
 			return checkRoot()
 		},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:  "source",
+				Usage: "Source for upgrade. Composed of `type:address`. Accepts `file:`,`dir:` or `oci:` for the type of source.\nFor example `file:/var/share/myimage.tar`, `dir:/tmp/extracted` or `oci:repo/image:tag`",
```
mauromorales, line 461: Maybe I'm mistaken to ask this but it feels to me like this should be a proper URI instead of our own way of doing this ... but maybe this is a better UX, not sure

main.go
jimmykarily: I wondered about `docker` too but that's how it was before: https://github.com/kairos-io/kairos-agent/pull/152/files/5b9d043aa0986ea31a3765932d85a5412fda49ac#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L115

Not sure if `docker` and `oci` are handled differently. I can try to fix this here but it's a change irrelevant to this PR (it looks innocent but you never know :D )

pkg/config/spec.go
```diff
@@ -722,3 +698,27 @@ func isMounted(config *Config, part *v1.Partition) (bool, error) {
...
+
+func setSourceSize(cfg *Config, spec *v1.UpgradeSpec) error {
+	var size int64
+	var err error
+
+	if spec.RecoveryUpgrade {
+		size, err = GetSourceSize(cfg, spec.Recovery.Source)
+	} else {
+		size, err = GetSourceSize(cfg, spec.Active.Source)
+	}
```
mauromorales, line 710: I think in this cases it's better to use the control flow to resturn a temp variable than to call the function on both branches

```suggestion
        var source whatverTypeSourceIs
        
	if spec.RecoveryUpgrade {
		source = spec.Recovery.Source
	} else {
		source = spec.Active.Source
	}
	
	size, err = GetSourceSize(cfg, source)
```

main.go
jimmykarily: actually, the comment might be wrong. [We use `url.Parse` to parse this value](https://github.com/kairos-io/kairos-agent/blob/1170576dcce4018a3dffddf500295c3ac3a64a46/pkg/types/v1/common.go#L93), so `oci://` is probably the right format.

pkg/config/spec.go
```diff
@@ -722,3 +698,27 @@ func isMounted(config *Config, part *v1.Partition) (bool, error) {
...
+	if err != nil {
+		return err
+	}
+
+	cfg.Logger.Infof("Setting image size to %dMb", size)
+	// On upgrade only the active or recovery will be upgraded, so we dont need to override passive
+	if spec.RecoveryUpgrade {
```
mauromorales, line 717: and here we do it again, so maybe it would be better if it's just done once at the beginning and then either spec.Recovery or spec.Active is handleded for the rest of the function

pkg/config/spec.go
jimmykarily: Actually our first take was to store `spec.Recovery` or `spec.Active` to a variable so that we can use it below when setting the Size too. But for some reason it didn't work (pointers in Go and such). I'm actually curious to understand why so I'll give it another shot.

main.go
mauromorales: yeah good question, for me the main issue is that we have two places which do the validation and that they don't match ... don't want to slow the release of the patch version so definitely ok with keeping it as is

internal/agent/upgrade.go
```diff
@@ -148,17 +95,133 @@ func determineUpgradeImage(version string) (string, error) {
...
+	}
+
+	// Set uri both for active and recovery because we don't know what we are
+	// actually upgrading. The "upgradeRecovery" is just the command line argument.
+	// The user might have set it to "true" in the kairos config. Since we don't
+	// have access to that yet, we just set both uri values which shouldn't matter
+	// anyway, the right one will be used later in the process.
+	if source != "" {
+		upgrade["upgrade"]["recovery-system"] = map[string]string{
+			"uri": source,
```
Itxaka, line 131:
```
if upgradeRecovery{
  entry = "recovery-system"
else {
  entry = "system"
}

upgrade["upgrade"][entry] = map[string]string{
			"uri": source,
}
```

in fact, it can be even easier if we just declare the entry var at the start with the `entry := "system"` value and on line 120 we override it when we enter the `if upgradeRecovery`

Then in here we just set the `upgrade["upgrade"][entry] = map[string]string{"uri": source}` if `source != ""`

main.go
```diff
@@ -385,13 +378,23 @@ This command is meant to be used from the boot GRUB menu, but can be also starte
 			&cli.BoolFlag{
 				Name: "shell",
 			},
+			&cli.StringFlag{
```
Itxaka, line 381: we can just declare this flag at the top and reuse it everywhere?

main.go
Itxaka: it is, as an URI just defines the protocal:address and we want this to be generic enough.

main.go
Itxaka: docker was the old one, its valid, but not recommended as it doesnt necessary has to be a docker artifact, oci artifacts are much more expansive thant docker :)

Itxaka (approved):

pkg/config/spec.go
```diff
@@ -265,30 +265,9 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
...
-				cfg.Logger.Infof("Setting image size to %dMb", size)
-				// On upgrade only the active or recovery will be upgraded, so we dont need to override passive
-				if recoveryUpgrade {
-					spec.Recovery.Size = uint(size)
-				} else {
-					spec.Active.Size = uint(size)
-				}
-			}
-		}
+	err = setSourceSize(cfg, spec)
```
jimmykarily, line 268: This is too early to run this. We haven't unmarshalled with yip yet thus spec.Recovery.Source (or spec.Active.Source) is not set yet.

main.go
jimmykarily: done

internal/agent/upgrade.go
jimmykarily: The problem described in the comment still remains. What if in the kairos config we have set `upgrade.recovery = true` ? We don't know this at this point because we haven't read the config yet so we can't simply rely on the command line value for upgradeRecovery because that's only one way to set this.

main.go
jimmykarily: but the way it is now, we would return an error if "docker:" was used. Should I change that?

pkg/config/spec.go
jimmykarily: Code has been refactored. This no longer applies.

pkg/config/spec.go
jimmykarily: refactored code to address this. Now we unmarshal the yaml before we calculate sizes.

pkg/config/spec.go
jimmykarily: Code is now refactored, this no longer applies.

main.go
jimmykarily: @Itxaka you mean "it is" as in "it is wrong" ? Should I change the text to have forward slashes too? e.g. `oci://`, `file://`, `dir://` etc?

internal/agent/upgrade.go
```diff
@@ -148,17 +95,133 @@ func determineUpgradeImage(version string) (string, error) {
 		Version: version,
 	})
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
 	if img != "" {
-		return img, nil
+		return nil, nil
```
mauromorales, line 102: is this really meant to return nil, nil?

mauromorales (approved):

main.go
Itxaka: it is valid and the way of doing it, otherwise the value cannot be obtained if we use `protocol://address`

Dont ask me why, it just doesnt work with the `://` properly. Maybe our parsing is incorrect or we are using it wrong: https://goplay.tools/snippet/AMNXoOVJr_6

Itxaka: In any case, I think this would be out of scope for this patch.

internal/agent/upgrade.go
jimmykarily: Oops! No this should have been : 

```
	if img != "" {
		return v1.NewSrcFromURI(img)
	}
```
fixed now

internal/agent/upgrade.go
jimmykarily: Discussed in a call, leaving as is.

main.go
jimmykarily: Yes, I agress, it's out of scope.

PR #153, opened 2023-09-27
Take symlinks into account for partition size

mauromorales:
This refactoring of the calculation of the source size takes into account the size of the linked files, which fixes the issue with not enough space reported in kairos-io/kairos#1841, however the result of the calculation is not correct because there are scenarios where multiple links point to the same source e.g. `bin/x` points to `usr/bin/x` and `usr/sbin/x` also points to `usr/bin/x`. The only way to have the exact value would be to keep track of the calculated files sizes so in case a link points to one which has previously been calculated it doesn't add it. If we want this solution I can open an issue but for the patch fix I think we should go with this.

fixes kairos-io/kairos#1841

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -19,6 +19,7 @@ package config
 import (
 	"fmt"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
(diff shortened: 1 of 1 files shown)


Itxaka (approved):
Good catch!

PR #165, opened 2023-10-17
Make the agent behave more like an agent

Itxaka:
This syncs with what we expect of the agent start as a service, something that runs in there and checks from time to time

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -3,7 +3,9 @@ package agent
 import (
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"time"
 
 	v1 "github.com/kairos-io/kairos-agent/v2/pkg/types/v1"
 
@@ -19,6 +21,17 @@ import (
 
 // Run starts the agent provider emitting the bootstrap event.
 func Run(opts ...Option) error {
+	// capture ctrl+c and exit cleanly
+	channel := make(chan os.Signal, 1)
+	signal.Notify(channel, os.Interrupt)
+	go func() {
+		for sig := range channel {
+			// sig is a ^C, handle it
+			fmt.Printf("Got %s signal, exiting\n", sig)
+			os.Exit(1)
+		}
+	}()
+
 	o := &Options{}
 	if err := o.Apply(opts...); err != nil {
 		return err
@@ -27,7 +40,7 @@ func Run(opts ...Option) error {
 	os.MkdirAll("/usr/local/.kairos", 0600) //nolint:errcheck
 
 	// Reads config
-	c, err := config.Scan(collector.Directories(o.Dir...))
+	c, err := config.Scan(collector.Directories(o.Dir...), collector.NoLogs)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


Itxaka:
Is this ok @mudler ? Before this, on alpine the service kept on restarting time and time again as it was kind of set as a service.

This makes it behave more like a service, but Im not sure of the implications of really calling the bootstrap event over and over?

Is this correct? Do we expect this behaviour to be like this?

Itxaka:
The thing with this is:

kairos-agent is managed as a one off service on alpine
this means that the service is supposed to start and go away
currently this makes the logging from the provider to catch the whole tty1 and ttyS0 and doesnt allow loging or anything
I was thinking of moving the agent to be more like a normal service
but on its current implementation it runs, and then exits which marks the service as failed and makes openrc restart it again and again
so I thougth about making it more service-like


currently exploring other ways of dealing with it.

Itxaka:
I think the real problem with this is:

kairos agent behaves differently
 - if no provider, it just runs and exists
 - if provider, then it waits for provider to end before exiting

Itxaka:
default inittab:

```
# /etc/inittab

::sysinit:/sbin/openrc sysinit
::sysinit:/sbin/openrc boot
::wait:/sbin/openrc default

# Set up a couple of getty's
tty1::respawn:/sbin/getty 38400 tty1
tty2::respawn:/sbin/getty 38400 tty2
tty3::respawn:/sbin/getty 38400 tty3
tty4::respawn:/sbin/getty 38400 tty4
tty5::respawn:/sbin/getty 38400 tty5
tty6::respawn:/sbin/getty 38400 tty6

# Put a getty on the serial port
#

# Stuff to do for the 3-finger salute
::ctrlaltdel:/sbin/reboot

# Stuff to do before rebooting
::shutdown:/sbin/openrc shutdown

ttyS0::respawn:/sbin/getty -L ttyS0 115200 vt100
````

Notice how the default runlevel has a `wait` on it, so it wont start the ttys until it has started everything

but the status shows that kairos-agent under a provider is marked as starting forever:

```
Runlevel: default
 cos-setup-network                                                                         [  started  ]
 edgevpn                                                                       [  started 00:11:39 (0) ]
 cos-setup-boot                                                                            [  started  ]
 kairos-agent                                                                              [ starting  ]
 cos-setup-reconcile                                                           [  started 00:11:40 (0) ]
Dynamic Runlevel: hotplugged
Dynamic Runlevel: needed/wanted
 hostname                                                                                  [  started  ]
 sysfs                                                                                     [  started  ]
 devfs                                                                                     [  started  ]
 modules                                                                                   [  started  ]
 hwclock                                                                                   [  started  ]
 fsck                                                                                      [  started  ]
 root                                                                                      [  started  ]
 localmount                                                                                [  started  ]
 dbus                                                                                      [  started  ]
```

**Comment by Itxaka** [comment]
well, adding a `&` in the service file also fixes this LMAO

File: internal/agent/agent.go [hunk 1]
```diff
@@ -19,6 +21,17 @@ import (
 
 // Run starts the agent provider emitting the bootstrap event.
 func Run(opts ...Option) error {
+	// capture ctrl+c and exit cleanly
+	channel := make(chan os.Signal, 1)
+	signal.Notify(channel, os.Interrupt)
+	go func() {
```
mudler, line 27: this will spawn up a goroutine and grow indefinetly on each time `Run` is called again

mudler (approved):
just a small nit but going in the good direction

PR #168, opened 2023-10-19
Keep track of visited files for size calculation

mauromorales:
The main change in this PR is to make the file calculation function aware of which files it has already visited, so they don't get added more than once.

For the test I decided not to validate the exact calculations because we add 100Mb at the end plus there's always a small value coming from what, I think, is the size of the directories and symlinks, but could be something else. I think it's better to just test what this feature introduces, which is the fact that the symlink doesn't count if the target file was counted already.

This PR was a mob session during our hacktoberfest event :tada: 

fixes kairos-io/kairos#1862


pkg/config/spec.go
```diff
@@ -562,45 +609,10 @@ func GetSourceSize(config *Config, source *v1.ImageSource) (int64, error) {
...
-					return err
-				}
-
-				size += fileInfo.Size()
-			}
+		filesVisited = make(map[string]bool)
 
-			return nil
+		err = fsutils.WalkDirFs(config.Fs, source.Value(), func(path string, d fs.DirEntry, err error) error {
+			return getDirSize(size, filesVisited, path, d, err)
```
jimmykarily, line 615: How can this function change the value of `size`? It's not a pointer, it's just [an int64](https://github.com/kairos-io/kairos-agent/pull/168/files#diff-53b724197bc14c3828ce6725322b892cb788860b0526184b63d7d114edb60941R600). What am I missing?

mauromorales: in this case size is an accumulator that gets passed. WalkDirFs will call the function every time for a new file updating the value

pkg/config/spec.go
```diff
@@ -545,6 +545,47 @@ func ReadUkiUpgradeFromConfig(c *Config) (*v1.UpgradeUkiSpec, error) {
...
+			return err
+		}
+
+		if !filepath.IsAbs(actualFilePath) {
+			// If it's a relative path, join it with the base directory path.
+			actualFilePath = filepath.Join(filepath.Dir(path), actualFilePath)
+		}
+	}
+
+	fileInfo, err := os.Stat(path)
```
jimmykarily, line 575:
```suggestion
	fileInfo, err := os.Stat(actualFilePath)
```

jimmykarily (approved):

pkg/config/spec.go
```diff
@@ -545,6 +545,47 @@ func ReadUkiUpgradeFromConfig(c *Config) (*v1.UpgradeUkiSpec, error) {
...
+	fileInfo, err := os.Stat(actualFilePath)
+	if os.IsNotExist(err) || fileList[actualFilePath] {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	*size += fileInfo.Size()
+	fileList[actualFilePath] = true
```
Itxaka, line 584: isnt this gonna be terrible for resource compsumption? Like we are really storing a full map of files with a bool? How does this look with a fs with 50000 files?

mauromorales: @Itxaka good question. So far, running it on an installation, does not seem to impact the speed, but I'm going to add some debugging to get an idea of how many files we are talking about, and I could use that number to initialize the map so that at least resizing does not need to happen as often

Itxaka: mostly thinking about things like rpi with smaller amounts of mem

jimmykarily: Maybe this helps:

```
package main

import (
	"fmt"
	"runtime"
	"strconv"
)

func main() {
	var mymap = map[string]bool{}

	for i := 0; true; i++ {
		mymap[strconv.Itoa(i)] = true
		if i%100 == 0 {
			fmt.Printf("i = %+v\n", i)
			PrintMemUsage()
		}
	}
}

func PrintMemUsage() {
	var m runtime.MemStats
	runtime.ReadMemStats(&m)
	fmt.Printf("Alloc = %v MiB", m.Alloc/1024/1024)
	fmt.Printf("\tTotalAlloc = %v MiB", m.TotalAlloc/1024/1024)
	fmt.Printf("\tSys = %v MiB", m.Sys/1024/1024)
	fmt.Printf("\tNumGC = %v\n", m.NumGC)
}

```

jimmykarily: PrintMemUsage copied from here: https://gophercoding.com/print-current-memory-usage/

jimmykarily: sample line:

```
i = 13661300
Alloc = 1121 MiB	TotalAlloc = 1493 MiB	Sys = 1361 MiB	NumGC = 9
```

13.5 million keys -> 1.5 Gb RAM ?

Itxaka: > sample line:
> 
> ```
> i = 13661300
> Alloc = 1121 MiB	TotalAlloc = 1493 MiB	Sys = 1361 MiB	NumGC = 9
> ```
> 
> 13.5 million keys -> 1.5 Gb RAM ?

so extrapolating from there, 1 million entries = 150Mb, 100k entries 15Mb ?

If that is accurate then it should be negligent to store 50k entries

mauromorales: On Ubuntu I see 27K files but I don't see an impact in memory usage at all:

`var filesVisited map[string]bool`
Alloc = 4 MiB
TotalAlloc = 8 MiB

`filesVisited = make(map[string]bool, 30000)`
Alloc = 4 MiB
TotalAlloc = 8 MiB

After appending all the values to the map
Alloc = 4 MiB
TotalAlloc = 8 MiB

PR #172, opened 2023-11-07
Identify android board model

Itxaka:


internal/agent/upgrade.go
```diff
@@ -65,11 +67,14 @@ func Upgrade(
 		return err
 	}
 
-	upgradeAction := action.NewUpgradeAction(c, upgradeSpec)
-
-	err = upgradeAction.Run()
-	if err != nil {
-		return err
+	if boardutils.GetAndroidBoardModel() == constants.QCS6490 {
```
mudler, line 70: what if we call it GenericAndroidModel for now? is there anything really specific of QCS?

Itxaka: I have zero idea, that's why I made it very specific for this board as others may not work the same

PR #173, opened 2023-11-14
qcs6490 upgrade path

Itxaka:


internal/agent/upgrade.go
```diff
@@ -60,16 +62,21 @@ func Upgrade(
...
+			return err
+		}
+	} else {
+		err = upgradeSpec.Sanitize()
+		if err != nil {
+			return err
+		}
+		if err = action.NewUpgradeAction(c, upgradeSpec).Run(); err != nil {
+			return err
+		}
```
mauromorales, line 79: I think it would be better to move this control flow inside NewUpgradeAction, or some wrapper around it

pkg/action/upgrade_qcs6490.go
```diff
@@ -0,0 +1,106 @@
+/*
+Copyright © 2022 SUSE LLC
```
mauromorales, line 2: Is this a suse or a kairos maintainers one?

internal/agent/upgrade.go
Itxaka: yeah problem is the types no? If newupgradeAction can return either a UpgradeAction or Upgradeqcs6490Action then we need to start asbtracting the types and casting it to the final type and such

I really dislike that in go :(

Itxaka: Also there is the issue with circular deps, there are some things that bring other things and other things which results in an endless circular dependency hell :sob:

pkg/types/v1/config.go
```diff
@@ -209,6 +209,7 @@ type Partition struct {
 	Size            uint     `yaml:"size,omitempty" mapstructure:"size"`
 	FS              string   `yaml:"fs,omitempty" mapstrcuture:"fs"`
 	Flags           []string `yaml:"flags,omitempty" mapstrcuture:"flags"`
+	Label           string
```
Itxaka, line 212: while this changes the struct, it should not affect anything, just stores the partitionLabel here which is info that we had but never used anywhere.

pkg/utils/boards/boards.go
```diff
@@ -0,0 +1,113 @@
...
+			return ""
+		}
+
+	}
+
+	return ""
+}
+
+// CheckDeps checks if the dependencies are installed on the system
+func CheckDeps(runner v1.Runner) error {
```
Itxaka, line 47: umm this could be dropped I think, I added it when I was unable to inject the deps in the base image directly but now I dont think it makes much sense as we only support upgrading from our "golden image" which has those deps in it

pkg/action/upgrade_qcs6490.go
Itxaka: This is a copypaste one xDDDD

pkg/config/spec.go
```diff
@@ -292,7 +292,7 @@ func setUpgradeSourceSize(cfg *Config, spec *v1.UpgradeSpec) error {
 		targetSpec = &(spec.Active)
 	}
 
-	if targetSpec.Source.IsEmpty() {
+	if targetSpec.Source == nil || targetSpec.Source.IsEmpty() {
```
Itxaka, line 295: this is extra but nice to have. Found out that if we reach here with an empty source, source will be nil and the agent will panic. So better to have that check in there. Should not happen in mainline as we got checks for that before on the spec.Sanitize()

pkg/utils/boards/boards.go
```diff
@@ -0,0 +1,99 @@
...
+
+// GetAndroidBoardModel returns the board model if the system is an Android board
+func GetAndroidBoardModel() string {
+	// Check if we are running on an Android board
+	if IsAnAndroidBoard() {
+		buildProp, err := godotenv.Read("/build.prop")
+		if err != nil {
+			return ""
+		}
+		switch buildProp["ro.product.board"] {
```
Itxaka, line 34: I made this generic, but could easily merge this and the above function into one to check just for this board. thoughts?

internal/agent/upgrade.go
mauromorales: I don't think the types should be a problem if we use generics ... but these changes can be done on a different PR, I don't want to block the merging

pkg/utils/boards/boards.go
mauromorales: I think this is much better than just checking for this board, that way if tomorrow we for any reason have another board, this is there and we avoid a quick and dirty :P

mauromorales (approved):

pkg/config/spec.go
mauromorales: maybe we should have a function for it with a comment, or just the comment, because your explanation makes it clear, but if I just read the code I wouldn't know this is the case

internal/agent/upgrade.go
Itxaka: noh, its mainly about circular deps between the actions, the utils, and the types pkg folders, I had a bad time ttrying to stop everything depending on everything :(

mauromorales: I see, yeah, if we have circular deps that's a bigger issue :/

pkg/config/spec.go
Itxaka: Good point, will add a comment

PR #174, opened 2023-11-20
Call `exec.Command` properly, splitting the cmd and args

jimmykarily:
This was failing but wasn't visible because it needs

`fail_on_bundles_errors: true` to be set in the Kairos config.

We need to find whether we need a new test or an existing one can be adapted to catch this.

### DIFF
--- a/internal/agent/hooks/bundles.go
+++ b/internal/agent/hooks/bundles.go
@@ -1,13 +1,14 @@
 package hook
 
 import (
+	"os"
+	"os/exec"
+	"syscall"
+
 	config "github.com/kairos-io/kairos-agent/v2/pkg/config"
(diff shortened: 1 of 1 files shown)


jimmykarily:
Adding `fail_on_bundles_errors: true` to [the config of the bundles test](https://github.com/kairos-io/kairos/blob/master/tests/assets/live-overlay.tmpl) would make the test fail but the way it fails is by a timeout waiting for a reboot to "active". This would take 30 minutes and would fail with no helpful error. The way the test runs, I'm not sure where the error is written, while it's there is one runs `kairos-agent manual-install` with the same config.

The test needs to be refactored to allow inspection of the installation progress with faster failure. We can then add the above option to make sure we don't break it again.

PR #178, opened 2023-12-05
Fix unwanted reboot from install config in upgrade/reset

Itxaka:
provided a call like this:

`elemental-agent --debug manual-install --poweroff --device "dev" --source oci:source /tmp`

it generates the proper collector config to be applied by the spec:

```
  Config: collector.Config{
    "install": collector.Config{
      "device": "dev",
      "poweroff": true,
      "reboot": false,
      "system": collector.Config{
        "uri": "oci:source",
      },
    },
  },

```

So it falls in line with what we been doing in other places, all config merged in a single place.

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -63,28 +63,16 @@ func ManualInstall(c, sourceImgURL, device string, reboot, poweroff, strictValid
 	}
 
 	cliConf := generateInstallConfForCLIArgs(sourceImgURL)
+	cliConfManualArgs := generateInstallConfForManualCLIArgs(device, reboot, poweroff)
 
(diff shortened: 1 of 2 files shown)


internal/agent/install.go
```diff
@@ -350,3 +338,18 @@ func generateInstallConfForCLIArgs(sourceImageURL string) string {
     uri: %s
 `, sourceImageURL)
 }
+
+// generateInstallConfForManualCLIArgs creates a kairos configuration for flags passed via manual install
+func generateInstallConfForManualCLIArgs(device string, reboot, poweroff bool) string {
+	cfg := fmt.Sprintf(`install:
```
mauromorales, line 344: don't we miss `#cloud-config` header here?

PR #186, opened 2023-12-18
UKI improvements

Itxaka:
- Check if OCI image exists in uki upgrade spec
 - Get efi partition from existing partitions
 - Calculate free space on uki partition
 - Remove uki files if they are over the limit during upgrade . Limit is currently 3 so if during upgrade we find 3 files, we remove one before copying the source.

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -18,6 +18,8 @@ package config
 
 import (
 	"fmt"
+	"github.com/google/go-containerregistry/pkg/crane"
+	"golang.org/x/sys/unix"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -538,6 +540,18 @@ func ReadUkiInstallSpecFromConfig(c *Config) (*v1.InstallUkiSpec, error) {
 func NewUkiUpgradeSpec(cfg *Config) (*v1.UpgradeUkiSpec, error) {
 	spec := &v1.UpgradeUkiSpec{}
 	err := unmarshallFullSpec(cfg, "upgrade", spec)
+	// TODO: Use this everywhere?
+	cfg.Logger.Infof("Checking if OCI image %s exists", spec.Active.Source.Value())
+	if spec.Active.Source.IsDocker() {
+		_, err := crane.Manifest(spec.Active.Source.Value())
+		if err != nil {
+			if strings.Contains(err.Error(), "MANIFEST_UNKNOWN") {
+				return nil, fmt.Errorf("oci image %s does not exist", spec.Active.Source.Value())
+			}
+			return nil, err
+		}
+	}
+
 	// Get the actual source size to calculate the image size and partitions size
 	size, err := GetSourceSize(cfg, spec.Active.Source)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/constants/constants.go
```diff
@@ -112,6 +112,7 @@ const (
 	UkiCdromSource    = "/run/install/cdrom"
 	UkiEfiDir         = "/efi"
 	UkiEfiDiskByLabel = `/dev/disk/by-label/` + EfiLabel
+	UkiMaxEntries     = 3
```
mauromorales, line 115: what does the max entries stand for?

pkg/uki/upgrade.go
```diff
@@ -34,19 +40,62 @@ func (i *UpgradeAction) Run() (err error) {
...
+
 	_ = elementalUtils.RunStage(i.cfg, "kairos-uki-upgrade.after")
 	_ = events.RunHookScript("/usr/bin/kairos-agent.uki.upgrade.after.hook") //nolint:errcheck
 
 	return hook.Run(*i.cfg, i.spec, hook.AfterUkiUpgrade...)
 }
+
+func (i *UpgradeAction) getEfiFiles() ([]string, error) {
+	var efiFiles []string
+	files, err := os.ReadDir(filepath.Join(i.spec.EfiPartition.MountPoint, "EFI", "kairos"))
```
mauromorales, line 90: I wonder if the function getEfiFile should belong to the UpgradeAction or to the MountPoint, I'd be more inclined to the latter seeing this here (it would then need to be exported, but still feels more likely to me that I would want the efi files from a mount point than from an upgrade action so to speak.

Itxaka: yes, I was struggling on where to put this here, also thougth about making it part of the fsutils somehow but didnt want to start playing with the imports here and there xD

Also notice that the EfiPartition is a very generic struct(v1.Partition) that works for everything else, so having it part of that felt a bit wrong as it was not generic enough as to be valid for anything else in the agent. Is very very specific to uki and to the upgrade part.

pkg/constants/constants.go
mauromorales: Makes sense, thanks for clarifying

pkg/uki/upgrade.go
```diff
@@ -34,19 +40,62 @@ func (i *UpgradeAction) Run() (err error) {
...
-	// TODO: Check size of source to see if we can upgrade
 	// TODO: Check number of existing UKI files
-	// TODO: Load them, order them via semver
-	// TODO: Remove the latest one if its over the max number of entries
+	efiFiles, err := i.getEfiFiles()
+	if err != nil {
+		return err
+	}
+	i.cfg.Logger.Infof("Found %d UKI files", len(efiFiles))
+	i.cfg.UkiMaxEntries = 1
```
Itxaka, line 49: shit, leftover of testing xDDD

mauromorales (approved):
lgtm

pkg/config/spec.go
```diff
@@ -538,6 +540,18 @@ func ReadUkiInstallSpecFromConfig(c *Config) (*v1.InstallUkiSpec, error) {
 func NewUkiUpgradeSpec(cfg *Config) (*v1.UpgradeUkiSpec, error) {
 	spec := &v1.UpgradeUkiSpec{}
 	err := unmarshallFullSpec(cfg, "upgrade", spec)
+	// TODO: Use this everywhere?
+	cfg.Logger.Infof("Checking if OCI image %s exists", spec.Active.Source.Value())
```
Itxaka, line 544: I tested this and it seems to work, I tihnk we should apply this everywhere, maybe when we construct the source if we find its a docker image we do the check to see if the tag exists. This is very nice because it would be done really soon, while on a install for example, a lot of things are done before we check the tag.

Could probabyl move to to sdk along the image utils.

Something similar for other sources can be done

PR #189, opened 2023-12-29
Copy grub+shim from the rootfs especific paths

Itxaka:
Its simpler to provide a package yes, but then we cnanot support secureboot in all distros that provide signed artifacts.

This should allow us to drop the system/grub2 package from the framework as the raw_images use other packages.

This also requires changes to enki to do basically the same and search the grub+shim used to boot the livecd under EFI in the rootfs as we do here

Requires: 
 - https://github.com/kairos-io/osbuilder/pull/107
 - https://github.com/kairos-io/enki/pull/17
 - https://github.com/kairos-io/packages/pull/607
 - https://github.com/kairos-io/kairos-framework/pull/50
 - This PR to be merged
 - New framework version https://github.com/kairos-io/kairos-framework/releases/tag/v2.6.0
 - New enki version https://github.com/kairos-io/enki/releases/tag/v0.0.4
 - New osbuilder version with the new enki version https://github.com/kairos-io/osbuilder/releases/tag/v0.11.0
 - Kairos PR to have the shim packages included on each flavor


pkg/utils/utils_test.go
```diff
@@ -840,7 +840,19 @@ var _ = Describe("Utils", Label("utils"), func() {
...
+				// Create fake shim
+				err = fsutils.MkdirAll(fs, filepath.Join(rootDir, "/usr/share/efi/x86_64/"), constants.DirPerm)
+				Expect(err).ShouldNot(HaveOccurred())
+				err = fs.WriteFile(filepath.Join(rootDir, "/usr/share/efi/x86_64/", constants.SignedShim), []byte(""), constants.FilePerm)
+				Expect(err).ShouldNot(HaveOccurred())
+				err = fs.WriteFile(filepath.Join(rootDir, "/usr/share/efi/x86_64/shim.efi"), []byte(""), constants.FilePerm)
+				Expect(err).ShouldNot(HaveOccurred())
+				grub = utils.NewGrub(config)
+				err = grub.Install(target, rootDir, bootDir, constants.GrubConf, "", true, "")
+				Expect(err).To(HaveOccurred())
```
jimmykarily, line 853: [minor] maybe use `MatchError` or similar to make sure we don't match random errors but the expected one

Itxaka: We match the error message below for this specific error that we want.

pkg/utils/grub.go
```diff
@@ -169,37 +170,89 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
...
-
+		// This seems like a chore while we could provide a package for those bundled files as they are just a shim and a grub efi
+		// BUT this is needed for secureboot
+		// The shim contains the signature from microsoft and the shim provider (i.e. upstream distro like fedora, suse, etc)
+		// So if we use the shim+grub from a generic package (i.e. ubuntu) it WILL boot with secureboot
+		// but when loading the kernel it will fail because the kernel is not signed by the shim provider or grub provider
+		// the kernel signature would be from fedora while the shim signature would be from ubuntu
+		// This is why if we want to support secureboot we need to copy the shim+grub from the rootfs default paths instead of
+		// providing a generic package
+		shimFiles := utils.GetEfiShimFiles(g.config.Arch)
```
mauromorales, line 181: I see some similitudes here with https://github.com/kairos-io/enki/pull/17/files

Like on the other PR, I would recommend to have different functions for the ship and grub processes. This way maybe we can see some commonalities that can be abstracted later on, but even if that's not the case I think it makes the code easier to read and review

Itxaka: yes, I wanted to test it in both places and maybe search for a way to centralize it on a helper function in the sdk so it doesnt diverge too much.

I will take a look and see if its possible, otherwise I will extract those things into a couple of functions

pkg/constants/constants.go
```diff
@@ -128,20 +127,6 @@ func GetDefaultSquashfsCompressionOptions() []string {
 	return []string{"-comp", "gzip"}
 }
 
-func GetGrubFilePaths(arch string) []string {
```
mudler, line 131: I'm concerned on the "legacy" portion of Kairos, but most importantly this mechanism was used also to protect from nuances of depending on the GRUB version shipped from the distro. It gives more control in what are you actually using.

I'm not against this change, but I'm saying we should fallback if we don't find an EFI file in the expected locations - so to say, let's have a `/usr/share/kairos/efi/*` if that's present takes precedence, otherwise we scan and if we don't find it, we just fail. That is also a good safety net meanwhile switching to use native packages wherever needed.

Itxaka: we have a fallback efi already but this is done when building the iso: https://github.com/kairos-io/enki/blob/main/pkg/action/build-iso.go#L267 and https://github.com/kairos-io/enki/blob/main/pkg/action/build-iso.go#L278

if we dont find efi files in the expected locations, we fall back to the generic one provided by the packages.

Itxaka: if we want secureboot this is the only way of providing it, by using artifacts directly from the distro owners. Its similar to the kernel part, we dont know the difference in kernels and that can screw us easily, but...we expect the upstream distros to have done their due diligence and make sure it works.

Otherwise there is no way of supporting secureboot unless we use our own keys to sign everything (including the kernel!)

Itxaka: > gotcha, I see, maybe we can start to make this more first-class in Kairos and have our own location (like we do for /etc/os-release variables for instance) easier to track and we don't have to overwrite distro-specific paths (that might be hard to realize in case of debug, totally not-obvious)

yes, this was my idea as well, provide them in a very specific place instead of all along, but this change has already so many changes across so many places that I didnt wanted to introduce yet another change to the packages without having this out and properly tested (the idea is to add a couple of tests in kairos to use EFI+secureboot once this is in)

mudler (approved):

PR #196, opened 2024-01-09
Run install off the same command

Itxaka:
Changes the uki install to be run with the same command as the normal install so it can leverage the normal workflow.

autodetects if is under uki by checking the sentinel set by immucore and the cmdline

Part of https://github.com/kairos-io/kairos/issues/2049

### DIFF
--- a/internal/agent/install.go
+++ b/internal/agent/install.go
@@ -20,7 +20,6 @@ import (
 	"github.com/kairos-io/kairos-agent/v2/internal/cmd"
 	"github.com/kairos-io/kairos-agent/v2/pkg/action"
 	"github.com/kairos-io/kairos-agent/v2/pkg/config"
-	v1 "github.com/kairos-io/kairos-agent/v2/pkg/types/v1"
 	events "github.com/kairos-io/kairos-sdk/bus"
 	"github.com/kairos-io/kairos-sdk/collector"
 	"github.com/kairos-io/kairos-sdk/machine"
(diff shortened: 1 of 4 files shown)


internal/agent/install.go
```diff
@@ -217,66 +217,90 @@ func RunInstall(c *config.Config) error {
...
 	}
 
-	// Load the installation spec from the Config
-	installSpec, err := config.ReadInstallSpecFromConfig(c)
-	if err != nil {
-		return err
-	}
+	// UKI path. Check if we are on UKI AND if we are running off a cd, otherwise it makes no sense to run the install
+	// From the installed system
+	if internalutils.IsUki() && internalutils.UkiBootMode() == internalutils.UkiRemovableMedia {
```
mauromorales, line 222: can we introduce two different functions for the control flow instead of putting it all in this function?

Itxaka: can you rephrase that? I dont get youo you are suggesting sorry. 

Also, this should be removed and only the uki boot mode one used as that internally cheks if its on uki first.

mauromorales: sorry my bad, I forget ppl are not in my head LOL

I felt that it would be better to have the uki and non-uki process put in different functions, because the outer function was already big enough and now it increased

if it will be removed soon then no need to change

mauromorales (approved):

internal/agent/install.go
Itxaka: ahh I see!

Will do

Itxaka: nicer?

jimmykarily (approved):

mauromorales (approved):

PR #198, opened 2024-01-10
Fix uki mode detection

Itxaka:


internal/agent/install.go
```diff
@@ -218,8 +218,13 @@ func RunInstall(c *config.Config) error {
 
 	// UKI path. Check if we are on UKI AND if we are running off a cd, otherwise it makes no sense to run the install
 	// From the installed system
-	if internalutils.UkiBootMode() == internalutils.UkiRemovableMedia {
-		return runInstallUki(c)
+	if internalutils.IsUki() {
+		c.Logger.Debugf("UKI mode: %s\n", internalutils.UkiBootMode())
```
Itxaka, line 222: bit more of logging to be able to check what mode are we in, in case we broke it like I already did :D

pkg/utils/common.go
```diff
@@ -520,7 +520,7 @@ const (
 func UkiBootMode() state.Boot {
 	if IsUki() {
 		_, err := os.Stat("/run/cos/uki_boot_mode")
-		if err != nil {
```
Itxaka, line 523: wops, I broke it on the first iteration by mixing the check :D

internal/agent/install.go
```diff
@@ -289,8 +294,9 @@ func dumpCCStringToFile(c *config.Config) (string, error) {
 		c.Logger.Error("Error creating temporary file for install config: %s\n", err.Error())
 		return "", err
 	}
-	defer os.RemoveAll(f.Name())
-
+	defer func(f *os.File) {
+		_ = f.Close()
```
jimmykarily, line 298: no cleanup?

jimmykarily (approved):

internal/agent/install.go
jimmykarily: closing the file here is correct. Removing it should be the responsibility of the caller. The temp directory is gone after installation right? Leaving it there is ok then.

PR #199, opened 2024-01-11
Add "--all" flag to list-releases to print older releases too

jimmykarily:
This will allow us to fix [the broken test in master](https://github.com/kairos-io/kairos/actions/runs/7479018843/job/20360820413) and also allow people to downgrade to older releases.


mauromorales (approved):
Niiiice :raised_hands:

PR #200, opened 2024-01-11
Remove cutSuffix as its go >=1.20 only

Itxaka:


Itxaka:
Tested and works as expected, merging directly to unblock building under fips

mauromorales (commented):
:+1:

PR #204, opened 2024-01-19
Restore partitions config from cloudconfig

Itxaka:
Now you can override the partition size from config as it used to be.

We also have a safeguard to avoid not having enough space for the image size, logging to mark what size we did end up with and if we had to revert to a safe size.

Fixes: https://github.com/kairos-io/kairos/issues/2151


mauromorales (approved):
lgtm

PR #205, opened 2024-01-19
Fix --source flag

Itxaka:
Seems like we were mainly ignoring the source flag from the start. We had a nice value in the struct to store the source coming from the config or flag but were setting the spec flags directly instead. The config should have gone into the config directly and then the spec Imagesrc is generated from that.

This should fix the takeover install and probably other use cases that we have not hit before as the --source flag seems to be rarely used, mostly it defaulted to the media dir which is the main deployment path (iso,netboot or usb)

Related to: https://github.com/kairos-io/kairos/issues/1528

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -69,10 +69,10 @@ func NewInstallSpec(cfg *Config) (*v1.InstallSpec, error) {
 	}
 	// Then any user provided source
 	if cfg.Install.Source != "" {
-		fmt.Println("Adding source!")
 		activeImg.Source, _ = v1.NewSrcFromURI(cfg.Install.Source)
 	}
 	// If we dont have any just an empty source so the sanitation fails
+	// TODO: Should we directly fail here if we got no source instead of waiting for the Sanitize() to fail?
 	if !isoRootExists && cfg.Install.Source == "" {
 		activeImg.Source = v1.NewEmptySrc()
 	}
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -69,11 +69,12 @@ func NewInstallSpec(cfg *Config) (*v1.InstallSpec, error) {
 	}
 	// Then any user provided source
 	if cfg.Install.Source != "" {
+		fmt.Println("Adding source!")
 		activeImg.Source, _ = v1.NewSrcFromURI(cfg.Install.Source)
 	}
 	// If we dont have any just an empty source so the sanitation fails
 	if !isoRootExists && cfg.Install.Source == "" {
-		return nil, fmt.Errorf("no installation source found")
+		activeImg.Source = v1.NewEmptySrc()
 	}
 
 	if recoveryExists {
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -63,10 +63,17 @@ func NewInstallSpec(cfg *Config) (*v1.InstallSpec, error) {
 	activeImg.FS = constants.LinuxImgFs
 	activeImg.MountPoint = constants.ActiveDir
 
+	// First try to use the install media source
 	if isoRootExists {
 		activeImg.Source = v1.NewDirSrc(constants.IsoBaseTree)
-	} else {
-		activeImg.Source = v1.NewEmptySrc()
+	}
+	// Then any user provided source
+	if cfg.Install.Source != "" {
+		activeImg.Source, _ = v1.NewSrcFromURI(cfg.Install.Source)
+	}
+	// If we dont have any just an empty source so the sanitation fails
+	if !isoRootExists && cfg.Install.Source == "" {
+		return nil, fmt.Errorf("no installation source found")
 	}
 
 	if recoveryExists {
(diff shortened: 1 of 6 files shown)


jimmykarily:
A test is broken. Also I wonder if we could add/change some test in the kairos repository to catch this.

Itxaka:
> A test is broken. Also I wonder if we could add/change some test in the kairos repository to catch this.

we should probably do a takeover install, but seems a bit complex. Boot with some random livecd, install docker, do takeover install, reboot and acceptance tests

maybe this has fixed the loop devices installation......which also stopped working around the same time...
that could be easier to test

Itxaka:
lol, this actually fixes the install to loop devices, so this is even better....we can create the raw images with kairos now...

```
qemu-img create test.img +20G
kairos-agent manual-install --device /dev/loop0 --source oci:quay.io/kairos/opensuse:leap-15.5-core-amd64-generic-v2.5.0 config.yaml
```

blkid of the loop devices after:
```
/dev/loop0p5: LABEL="COS_PERSISTENT" UUID="7c98afd2-435c-4b3b-81f4-ea5757749270" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="persistent" PARTUUID="9324d87d-83f1-4563-a21e-14965d680040"
/dev/loop0p3: LABEL="COS_RECOVERY" UUID="dd9e6401-4ffd-491e-9d01-efdf825a0cef" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="recovery" PARTUUID="8d076e21-5203-4e12-86aa-319c1bbe3720"
/dev/loop0p1: SEC_TYPE="msdos" LABEL_FATBOOT="COS_GRUB" LABEL="COS_GRUB" UUID="3251-13C3" BLOCK_SIZE="512" TYPE="vfat" PARTLABEL="efi" PARTUUID="67a09af7-8919-40c1-b336-c0b2705149c1"
/dev/loop0p4: LABEL="COS_STATE" UUID="5907cbdf-22e9-420b-aa57-67c30bbcbe65" BLOCK_SIZE="4096" TYPE="ext4" PARTLABEL="state" PARTUUID="bfb89772-da52-44f9-81e7-9a8a2c039e39"
/dev/loop0p2: LABEL="COS_OEM" UUID="057f74e4-8304-4f35-83bd-5577003d2a3a" BLOCK_SIZE="1024" TYPE="ext4" PARTLABEL="oem" PARTUUID="7fc730c7-dfeb-4bb6-95f5-0b3db0b02a29"
```

so now the test.img is ready to be burn to a disk and booted...

jimmykarily:
> ```
> qemu-img create test.img +20G
> kairos-agent manual-install --device /dev/loop0 --source oci:quay.io/kairos/opensuse:leap-15.5-core-amd64-generic-v2.5.0 config.yaml
> ```

Some steps are missing here, no? How does the second command know about the `test.img` file created by the command above? Shouldn't we first mount the img?

If this is working, does it mean we can get rid of these 2 scripts?

- https://github.com/kairos-io/osbuilder/blob/master/tools-image/build-arm-image.sh
- https://github.com/kairos-io/osbuilder/blob/master/tools-image/prepare_arm_images.sh

Itxaka:
> > ```
> > qemu-img create test.img +20G
> > kairos-agent manual-install --device /dev/loop0 --source oci:quay.io/kairos/opensuse:leap-15.5-core-amd64-generic-v2.5.0 config.yaml
> > ```
> 
> Some steps are missing here, no? How does the second command know about the `test.img` file created by the command above? Shouldn't we first mount the img?
> 
> If this is working, does it mean we can get rid of these 2 scripts?
> 
> - https://github.com/kairos-io/osbuilder/blob/master/tools-image/build-arm-image.sh
> - https://github.com/kairos-io/osbuilder/blob/master/tools-image/prepare_arm_images.sh
> 
> 
> 

Yes, sorry, mounting the img as a loop device is missing.


Potentially yes, we may be able to remove those scripts to generate raw images for arm64. RPI still won't work as it would need extra packages but we can probably work something out to generate RPI images directly

pkg/config/spec.go
```diff
@@ -63,9 +63,17 @@ func NewInstallSpec(cfg *Config) (*v1.InstallSpec, error) {
 	activeImg.FS = constants.LinuxImgFs
 	activeImg.MountPoint = constants.ActiveDir
 
+	// First try to use the install media source
 	if isoRootExists {
 		activeImg.Source = v1.NewDirSrc(constants.IsoBaseTree)
-	} else {
+	}
+	// Then any user provided source
+	if cfg.Install.Source != "" {
```
jimmykarily, line 71: Shouldn't this be tried before `isoRootExists` ? If the user explicitly specified a source image, shouldn't we just use that?

Itxaka: it goes in descending order and overwrites it if found to maintain what we used to did.

So:
 - set to isobasetree if it exists
 - overwrite with user provided source if user provided it

This way, if we dont provide the source it still falls back to the isotree as it did before
And then if neither of those are available it creates an empty source as it did before.

If we did the other way, the isorootree woul overwrite the user provided source and we dont want that.

jimmykarily: oh ok, I didn't expand all the code. The order just seemed wrong. All good.

jimmykarily (approved):

PR #212, opened 2024-01-26
Do not show devices that cant be used

Itxaka:
during interactive install all devices will appear but some of those are not desirable as they cannot be used to install.

Skip those in the list

### DIFF
--- a/internal/agent/interactive_install.go
+++ b/internal/agent/interactive_install.go
@@ -149,6 +149,10 @@ func InteractiveInstall(debug, spawnShell bool, sourceImgURL string) error {
 	block, err := ghw.Block()
 	if err == nil {
 		for _, disk := range block.Disks {
+			// skip useless devices (/dev/ram, /dev/loop, /dev/sr, /dev/zram)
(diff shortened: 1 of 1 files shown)


internal/agent/interactive_install.go
```diff
@@ -149,6 +149,10 @@ func InteractiveInstall(debug, spawnShell bool, sourceImgURL string) error {
 	block, err := ghw.Block()
 	if err == nil {
 		for _, disk := range block.Disks {
+			// skip useless devices (/dev/ram, /dev/loop, /dev/sr, /dev/zram)
+			if strings.HasPrefix(disk.Name, "loop") || strings.HasPrefix(disk.Name, "ram") || strings.HasPrefix(disk.Name, "sr") || strings.HasPrefix(disk.Name, "zram") {
```
mauromorales, line 153: Prefix, or suffix? ... Maybe I'm mistaken in assuming `disk.Name` has `/dev/` in it

mauromorales: ah nice, thanks!

PR #213, opened 2024-01-26
Remove entries on install

Itxaka:


pkg/uki/install.go
```diff
@@ -120,6 +124,52 @@ func (i *InstallAction) Run() (err error) {
...
+						return err
+					}
+					// Then remove the conf file
+					i.cfg.Logger.Debugf("Removing conf file %s", path)
+					err = i.cfg.Fs.Remove(path)
+					if err != nil {
+						return err
+					}
+					// Do not continue checking the conf file, we already done all we needed
+					break
```
jimmykarily, line 166: Doesn't this mean, we always remove one entry only? This `break` breaks the for loop in case a match is found. What if there was another match?

Itxaka: This breaks the inner loop of checking the conf against the entries, not the walk dir function, no?

Itxaka: yes, this is the inner loop, so once we match and remove the entry we skip going over the list of skip entries and continue the outer loop of the walkfs

mauromorales (approved):

pkg/uki/install.go
jimmykarily: oh true, my bad

jimmykarily (approved):

pkg/uki/install.go
Itxaka: umm maybe this one does exit the full loop somehow lmao

Itxaka:
now it works :D

PR #214, opened 2024-01-26
Set proper partition size based on the source

Itxaka:
When mounting the cdrom, we can calculate the partition size properly to be size of the source * 3

If we cannot infer the size, we just leave the default (3072 * 3)

Should be good enough for at least 8 different UKI files of 1gb each with the default.

Notice that this kicks in even if we are then removing efi files due to the skip entries to remove install entries, so there is a chance that we are going a bit over the size needed, but knowing that this has to survive the whole lifecycle from install to machine decommision, I would even bump this to be something big like 20/30gb hardcoded.


pkg/constants/constants.go
```diff
@@ -50,7 +50,7 @@ const (
 	OEMSize                 = uint(64)
 	PersistentSize          = uint(0)
 	BiosSize                = uint(1)
-	ImgSize                 = uint(3072)
+	ImgSize                 = uint(1024)
```
mauromorales, line 53: This is not used anywhere else?

mauromorales (dismissed):
lgtm

pkg/constants/constants.go
Itxaka: Oooh shit, good catch

Itxaka:
tested and works as expected

mauromorales (approved):

PR #216, opened 2024-01-30
[uki] Wait for oem to appear after encryption

Itxaka:
Wait a bit for oem to appear after it has been encrypted and unlocked before trying to mount it, otherwise if we are too fast and the disk is a bit slow or the kernel has not reported the changes yet, we hit an error becuase its not ready yet.

Also adds a close for the partitions when we finish dealing with them, otherwise we leave them open and unlocked

### DIFF
--- a/internal/agent/hooks/kcrypt_uki.go
+++ b/internal/agent/hooks/kcrypt_uki.go
@@ -1,6 +1,7 @@
 package hook
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


internal/agent/hooks/kcrypt_uki.go
```diff
@@ -96,10 +97,40 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
...
+		}
+	}()
+
+	// Here it can take the oem partition a bit of time to appear after unlocking so we need to retry a couple of time with some waiting
+	for i := 0; i < 10; i++ {
+		c.Logger.Infof("Waiting for unlocked partition to appear")
+		part, _ := utils.SH("blkid -L COS_OEM")
+		if part == "" {
+			c.Logger.Infof("Partition not found, waiting 1 second before retrying")
+			time.Sleep(5 * time.Second)
```
mauromorales, line 128: how about an incremental wait? so it does something like 1 sec, 5 secs, 15 secs or so ?

internal/agent/hooks/kcrypt_uki.go
```diff
@@ -96,10 +97,40 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
...
+			}
+		}
+	}()
+
+	// Here it can take the oem partition a bit of time to appear after unlocking so we need to retry a couple of time with some waiting
+	for i := 0; i < 10; i++ {
+		c.Logger.Infof("Waiting for unlocked partition to appear")
+		part, _ := utils.SH("blkid -L COS_OEM")
+		if part == "" {
+			c.Logger.Infof("Partition not found, waiting 1 second before retrying")
```
mauromorales, line 127: currently is set to 5 seconds ... maybe make it a variable and use it both for print and for waiting

internal/agent/hooks/kcrypt_uki.go
Itxaka: good point!

internal/agent/hooks/kcrypt_uki.go
Itxaka: done!

mauromorales (approved):
sweeet

internal/agent/hooks/kcrypt_uki.go
```diff
@@ -96,10 +97,40 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
...
+		_, _ = utils.SH("sync")
+		part, _ := utils.SH("blkid -L COS_OEM")
+		if part == "" {
+			c.Logger.Infof("Partition not found, waiting %d seconds before retrying", i)
+			time.Sleep(time.Duration(i) * time.Second)
+			continue
+		}
+		c.Logger.Infof("Partition found, continuing")
+		break
+	}
```
jimmykarily, line 133: We are being optimistic here and assume that we will always find the partition after 10 tries, because we don't handle the other case at all. Compared to the previous code, this at least tries more but maybe we could error out if the partition is not found after all the retries.

(I'm late to the party, the PR is closed, but if you think it's important, we can fix it in another PR)

Itxaka: I think this cannot fail in any other case, like this is only finding it after creating, if it fails it should fail before during the creation part, but if that succeeded there is no way the device is not showing up. This is just basically waiting for the kernel to refresh the devices and even then it happens on a very small occasion because we are too fast checking it out

PR #217, opened 2024-01-31
fix(source): initialize sources of images

mudler:
Should fix the panic in https://github.com/kairos-io/kairos/issues/2203

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -18,13 +18,14 @@ package config
 
 import (
 	"fmt"
-	"github.com/google/go-containerregistry/pkg/crane"
-	"golang.org/x/sys/unix"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"reflect"
 	"strings"
+
+	"github.com/google/go-containerregistry/pkg/crane"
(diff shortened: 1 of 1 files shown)


pkg/config/spec.go
```diff
@@ -221,6 +222,10 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
 	var recLabel, recFs, recMnt string
 	var active, passive, recovery v1.Image
 
+	active.Source = v1.NewEmptySrc()
```
Itxaka, line 225: this only affect normal upgrades, which dont have any issues with nil access :D

Fix should og unto `NewUpgradeUkiSpec`

mudler: right indeed.. I've just followed the panic here, did realize afterwards that we were actually skipping the UKI upgrade command and going into normal mode

jimmykarily:
This didn't fix it. We provided the actual fix elsewhere.

Itxaka:
Fixed at https://github.com/kairos-io/kairos-agent/pull/219

PR #221, opened 2024-02-02
Add reset for uki

Itxaka:


pkg/types/v1/config.go
```diff
@@ -541,8 +541,11 @@ func (i *UpgradeUkiSpec) ShouldReboot() bool   { return i.Reboot }
 func (i *UpgradeUkiSpec) ShouldShutdown() bool { return i.PowerOff }
 
 type ResetUkiSpec struct {
-	Reboot   bool `yaml:"reboot,omitempty" mapstructure:"reboot"`
-	PowerOff bool `yaml:"poweroff,omitempty" mapstructure:"poweroff"`
+	FormatPersistent bool `yaml:"reset-persistent,omitempty" mapstructure:"reset-persistent"`
```
mauromorales, line 544: again probably lack of knowledge on the whole installation so bear with me ... same applies for any other comment here so I don't repeat myself 🐼 

why do we use different naming here? it's the first time I read about Format, reset on the other hand is quite common to find.

internal/agent/hooks/hook.go
```diff
@@ -36,10 +36,14 @@ var AfterUkiInstall = []Interface{
 }
 
 // AfterUkiReset sets which Hooks to run after uki runs the install action
-var AfterUkiReset = []Interface{}
+var AfterUkiReset = []Interface{
```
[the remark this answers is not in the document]
Itxaka: yes, on a first try when implementing install, I did reuse the normal hooks, but it turned out that they were pretty much different and the hooks that we needed were also different so I just separated all of them to make it clearer which ones are which.

but indeed we could just use the same ones but do the check inside the hook somehow. Note that when this was first introduced the hook did not have access to the Config struct or Spec so they were much more limited, no they do have access to both so it should be easier to check.

But IMHO, for the start of this, I rather have them separated so we can deal with any issues that arise or anything that we need to add in a much more simpler way, and then we can merged them. As this is in internal API, there is no external breakage if we want to change this :D

internal/agent/hooks/hook.go
```diff
@@ -36,10 +36,14 @@ var AfterUkiInstall = []Interface{
...
+	&Lifecycle{},
+}
 
 // AfterUkiUpgrade sets which Hooks to run after uki runs the install action
-var AfterUkiUpgrade = []Interface{}
+var AfterUkiUpgrade = []Interface{
+	&Lifecycle{},
```
Itxaka, line 45: this was forgotten, so uki upgrade would not reset? have to check this further becuase I think we are doing this manually in the upgrade action which is bad

main.go
```diff
@@ -564,15 +564,20 @@ This command is meant to be used from the boot GRUB menu, but can likely be used
 				Name:  "unattended",
 				Usage: "Do not wait for user input and provide ttys after reset. Also sets the fast mode (do not wait 60 seconds before reset)",
 			},
+			&cli.BoolFlag{
```
Itxaka, line 567: this is a new flag to expose the underlying config. We allowed this via config or event but not via manual flags..

pkg/elemental/elemental.go
```diff
@@ -43,7 +43,14 @@ func NewElemental(config *agentConfig.Config) *Elemental {
 
 // FormatPartition will format an already existing partition
 func (e *Elemental) FormatPartition(part *v1.Partition, opts ...string) error {
-	e.config.Logger.Infof("Formatting '%s' partition", part.Name)
+	var name string
```
Itxaka, line 46: this is just UX, the DeviceMapper partition info is gathered via udev info. For luks devices the name does not appear anywhere so you would get an ugly message for those with no name `Formatting '' partition` 

This does a fallback in case its empty so at least the log message works and user is informed and doesnt look like we screwed up something.

internal/agent/hooks/hook.go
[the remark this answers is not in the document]
mauromorales: gotcha, thanks for explaining .... and yes I guess to get this out of the door faster and still stable that makes sense. cool

internal/agent/reset.go
```diff
@@ -20,13 +22,78 @@ import (
...
+	}
+
+	err = resetSpec.Sanitize()
+	if err != nil {
+		return err
+	}
+
+	resetAction := action.NewResetAction(cfg, resetSpec)
+	if err = resetAction.Run(); err != nil {
+		cfg.Logger.Errorf("failed to reset uki: %s", err)
```
mauromorales, line 54: I guess this one is not uki ... which relates to the point I was going to make ... reset and reset uki are so similar that it might make sense just to distinguish between readresetuki and readreset to avoid having disparity between these two functions

pkg/types/v1/config.go
Itxaka: This is unfortunately coming from the normal reset where we called it reset-oem/reset-persistent and needs to be kept for the configs and such.
But the real action that we are doing is formatting the partition. reset means nothing for persistent/oem as we just format the partition, so I though it may be clearer to have it like that so it also syncvs with the normal reset in which we have the same values.

I mean, `if FormatPersistent { FormatPartition(spec.Partitions.Persistent) }` seems nicer than `if ResetPersistent { FormatPartition(spec.Partitions.Persistent) }` as the second TO ME looks like more things are gonna be done to that persistent thing to "reset" it :rofl: 

But it could be me, I dont mind changing it.

internal/agent/reset.go
```diff
@@ -71,49 +139,51 @@ func Reset(reboot, unattended bool, dir ...string) error {
 
 	bus.Manager.Publish(sdk.EventBeforeReset, sdk.EventPayload{}) //nolint:errcheck
 
-	c, err := config.Scan(collector.Directories(dir...))
-	if err != nil {
-		return err
+	// Prepare a config from the cli flags
+	r := map[string]map[string]interface{}{
```
mauromorales, line 143: I think a struct would make more sense, to me it feels a bit strange to be doing too much stuff like this: `r["reset"]["reset-oem"]` and maybe I'm mistaken but then it would be easier to deduce because this would be automatic: `c.Reset.ResetOem = resetOem`

internal/agent/reset.go
Itxaka: yeah the problem with that is the fucking ugliness of the type casting shit we have to do.


Firs we would need to create an interface for ResetSpec that its more generic, so a ReadReset generic function could return a generic ResetSpec, then we have to cast that crap into the proper type and plead to ged that its not broken.

I dont known, I hate this metaprogramming or what the hell the name is on golang, its just not well done and involves a lot of ugly code.

I did something similar for the InstallSpec and the extra code that I had to add just for that was not worth it in my opinion: https://github.com/kairos-io/kairos-agent/blob/main/pkg/types/v1/config.go#L45

```

func (i *InstallSpec) ShouldReboot() bool                 { return i.Reboot }
func (i *InstallSpec) ShouldShutdown() bool               { return i.PowerOff }
func (i *InstallSpec) GetTarget() string                  { return i.Target }
func (i *InstallSpec) GetPartTable() string               { return i.PartTable }
func (i *InstallSpec) GetPartitions() ElementalPartitions { return i.Partitions }
func (i *InstallSpec) GetExtraPartitions() PartitionList  { return i.ExtraPartitions }

```

Madness.

Itxaka: Also this already goes down the rabbit whole into a more generic function that does the same for all specs `ReadSpecFromCloudConfig` which means they are all centralized and they all do the same thing.

Its just that on the agent the ugliness is hidden and a much clear view is shown.

pkg/types/v1/config.go
mauromorales: I like your approach, makes total sense if this is the actual action that gets executed, it's unfortunate it will be confusing but I don't think it's worth carrying the bad name over. And I'd say we should try to keep track of these things to long-term switch `reset-persisten` to `format-persistent` and deprecate the old names (same for any other case like this)

internal/agent/reset.go
mauromorales: oki, you have more experience with it so not a strong push from my side :P ... then the only change needed is on the error message to not say uki

internal/agent/reset.go
Itxaka: This is a good point, but this is then marshalled into a string to be passed to the collector, so I'm not sure how would it fit into that as it needs to be a reader for the collector to read it as another source.

mauromorales (approved):

internal/agent/reset.go
Itxaka: well, that was simpler that I expected LMAO

internal/agent/reset.go
```diff
@@ -71,49 +139,59 @@ func Reset(reboot, unattended bool, dir ...string) error {
...
 
-	bus.Manager.Publish(sdk.EventAfterReset, sdk.EventPayload{}) //nolint:errcheck
+	utils.SetEnv(c.Env)
+
+	return c, nil
+}
 
-	return hook.Run(*c, resetSpec, hook.AfterReset...)
+// ExtraConfigReset is the struct that holds the reset options that come from flags and events
+type ExtraConfigReset struct {
```
Itxaka, line 191: we should probably do the same on the upgrade.

I wonder if there may be any issues with overridable values...like we set some of this to false on the struct creation so maybe we are overriding values from the config with this ones....

Itxaka: cc @mauromorales

internal/agent/reset.go
```diff
@@ -140,26 +140,25 @@ func sharedReset(reboot, unattended, resetOem bool, dir ...string) (c *config.Co
...
+	r.Reset.ResetOem = resetOem
 
 	if resetOem {
-		r["reset"]["reset-oem"] = "true"
+		r.Reset.ResetOem = true
 	}
 
 	if reboot {
-		r["reset"]["reboot"] = reboot
+		r.Reset.Reboot = reboot
```
mauromorales, line 151: wait I just noticed, left side is boolean and right side is string, so probably also needs something like in 147 or like in 158

Itxaka: ? reboot is boolean no?

internal/agent/hooks/hook.go
Itxaka: yes, indeed, we were manually rebooting/poweroff-ing, probably becuase the lifecycle didnt hit

Fixed this plus got the hook together @mauromorales

internal/agent/reset.go
mauromorales: doesn't it come from the argument list in line 92? but indeed with the change it should work eitherway

internal/agent/reset.go
mauromorales: yeah I see your point, I would say we need to make sure we are setting them all every single time, and if we want to be extra careful then the logic to set them can go in one function that takes all 3

PR #223, opened 2024-02-07
Add default system config dir

mauromorales:
Fixes kairos-io/kairos#2221

### DIFF
--- a/main.go
+++ b/main.go
@@ -36,7 +36,7 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-var configScanDir = []string{"/oem", "/usr/local/cloud-config", "/run/initramfs/live"}
+var configScanDir = []string{"/oem", "/usr/local/cloud-config", "/run/initramfs/live", "/etc/kairos"}
 
 // ReleasesToOutput gets a semver.Collection and outputs it in the given format
 // Only used here.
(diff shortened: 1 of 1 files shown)


jimmykarily (commented):
We will need to document this. Here maybe? https://kairos.io/docs/reference/build-from-scratch/ not sure.

Itxaka:
I dont understand this. Are we just introducing an extra dir? What does this gives us that /system/oem or /oem doesnt?

jimmykarily:
> I dont understand this. Are we just introducing an extra dir? What does this gives us that /system/oem or /oem doesnt?

as far as I understand, this new directory is meant to be used by "vendors" to ship config along with the released images. The other paths are there for the users to place their own configs. I'm not sure why the vendors can't write to these directories. Probably because of how we mount them to persistent? (overlaying on top of the existing path)

Itxaka (approved):

mauromorales:
> /etc/elemental/config.yaml was supported as a way to inject default system configs. As part of those we could configure branding menu to be applied during upgrade and reset, and additional cloud config path to look into the system.

> Now we expect all that data to come from the user-data, however, we don't have a way to specify those settings when we build the images. The config collector scans configurations exclusively in paths that are either persistent or not accessible on build time

PR #224, opened 2024-02-08
Read /system/oem config

mauromorales:
Fixes kairos-io/kairos#2223

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,7 @@ import (
 
 var configScanDir = []string{
 	"/oem",
+	"/system/oem",
 	"/usr/local/cloud-config",
 	"/run/initramfs/live",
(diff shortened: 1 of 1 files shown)


mudler (approved):

main.go
```diff
@@ -38,6 +38,7 @@ import (
 
 var configScanDir = []string{
 	"/oem",
+	"/system/oem",
```
jimmykarily, line 41: I wonder what the relation with these paths is: https://github.com/kairos-io/immucore/blob/a2874ca3ee5385a0e81a9e9f8c95633939877656/internal/constants/constants.go#L11

If they are supposed to be the same paths, maybe it's worth moving them to the kairos-sdk and use them from there to make sure they never diverge.

mauromorales: good call, wasn't aware of these

PR #225, opened 2024-02-08
Bring uki install up to par with immucore

Itxaka:
Immucore now
(https://github.com/kairos-io/immucore/commit/a2874ca3ee5385a0e81a9e9f8c95633939877656) mounts the livecd automatically under the same route that its available on non-uki installs, so its a bit easier on our side as we can assume that the cdrom is mounted in the same path and use that.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -108,7 +108,7 @@ const (
 	Rsync      = "rsync"
 
 	UkiSource         = "/run/install/uki"
-	UkiCdromSource    = "/run/install/cdrom"
+	UkiCdromSource    = "/run/initramfs/live"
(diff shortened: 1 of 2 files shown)


mauromorales (approved):

pkg/constants/constants.go
```diff
@@ -108,7 +108,7 @@ const (
 	Rsync      = "rsync"
 
 	UkiSource         = "/run/install/uki"
-	UkiCdromSource    = "/run/install/cdrom"
+	UkiCdromSource    = "/run/initramfs/live"
```
jimmykarily, line 111: Couldn't we have used `LiveDir` which has the same value?

PR #226, opened 2024-02-09
fix(mount): check if path is mounted and create /run/initramfs/live if doesn't exist

mudler:
Attempt to fix https://github.com/kairos-io/kairos/issues/2224

Closes https://github.com/kairos-io/kairos/issues/2224.

### DIFF
--- a/pkg/uki/install.go
+++ b/pkg/uki/install.go
@@ -1,6 +1,7 @@
 package uki
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"strings"
@@ -35,8 +36,13 @@ func (i *InstallAction) Run() (err error) {
 
 	// immucore mounts CDROM under this path
 	_, err = i.cfg.Fs.Stat(constants.UkiCdromSource)
-	if err != nil {
-		i.cfg.Logger.Errorf("Error checking if %s exists: %s", constants.UkiCdromSource, err)
+	if errors.Is(err, os.ErrNotExist) {
+		// The directory is under /run, so we can create it if not present
+		if err := fsutils.MkdirAll(i.cfg.Fs, constants.UkiCdromSource, os.ModeDir|os.ModePerm); err != nil {
+			i.cfg.Logger.Errorf("Error creating %s: %s", constants.UkiCdromSource, err)
+		}
+	} else if err != nil {
+		i.cfg.Logger.Errorf("stat error for '%s'. error: %s", constants.UkiCdromSource, err)
 		return err
 	}
 	// Get source (from spec?)
@@ -50,10 +56,14 @@ func (i *InstallAction) Run() (err error) {
 		MountPoint: constants.UkiSource,
 	}
 
-	err = e.MountImage(image)
-	if err != nil {
-		return err
+	mounted, err := i.cfg.Mounter.IsMountPoint(constants.UkiCdromSource)
+	if !mounted || err != nil {
+		err = e.MountImage(image)
+		if err != nil {
+			return err
+		}
 	}
+
 	cleanup.Push(func() error {
 		return e.UnmountImage(image)
 	})


pkg/utils/partitions/getpartitions.go
```diff
@@ -209,3 +209,24 @@ func GetPartitionViaDM(fs v1.FS, label string) *v1.Partition {
 	}
 	return part
 }
+
+func IsPathMounted(fs v1.FS, s string) (bool, error) {
```
Itxaka, line 213: dont we have a util from the mount lib that checks this? Maybe we dont need to reimplement it already?

Itxaka: yep, there is a `IsNotMountPoint` from the mounter, maybe we can reuse it?

```
// IsNotMountPoint determines if a directory is a mountpoint.
// It should return ErrNotExist when the directory does not exist.
// IsNotMountPoint is more expensive than IsLikelyNotMountPoint.
// IsNotMountPoint detects bind mounts in linux.
// IsNotMountPoint enumerates all the mountpoints using List() and
// the list of mountpoints may be large, then it uses
// isMountPointMatch to evaluate whether the directory is a mountpoint.
func IsNotMountPoint(mounter Interface, file string) (bool, error) {
```

Itxaka: You also got `i.cfg.Mounter.IsLikelyNotMountPoint(constants.UkiCdromSource)` available but we are using a fakeMounter so we need to set a proper mounter on config build for that to work.

Whatever you wish, but thats the mounter check from kubernetes so I would say that has been thoroughly tested lol

Itxaka (approved):

pkg/utils/partitions/getpartitions.go
mudler: > You also got `i.cfg.Mounter.IsLikelyNotMountPoint(constants.UkiCdromSource)` available but we are using a fakeMounter so we need to set a proper mounter on config build for that to work.
> 
> Whatever you wish, but thats the mounter check from kubernetes so I would say that has been thoroughly tested lol

alright alright, moving to using IsMountPoint() then :)

pkg/uki/install.go
```diff
@@ -50,10 +56,14 @@ func (i *InstallAction) Run() (err error) {
 		MountPoint: constants.UkiSource,
 	}
 
-	err = e.MountImage(image)
-	if err != nil {
-		return err
+	mounted, err := i.cfg.Mounter.IsMountPoint(constants.UkiCdromSource)
```
Itxaka, line 59: argh shit, sorry this wont work. The mounter in here is a FakeMounter as we didnt needed a real mounter, soi this check is done in memory mounts only, not real ones....

Itxaka: or not, fuck let me check Im a bit confused today

Itxaka: yeah Im stupid, this is good, I was confusing this with enki.

pkg/uki/install.go
```diff
@@ -35,8 +36,13 @@ func (i *InstallAction) Run() (err error) {
 
 	// immucore mounts CDROM under this path
 	_, err = i.cfg.Fs.Stat(constants.UkiCdromSource)
-	if err != nil {
-		i.cfg.Logger.Errorf("Error checking if %s exists: %s", constants.UkiCdromSource, err)
+	if errors.Is(err, os.ErrNotExist) {
+		// The directory is under /run, so we can create it if not present
+		if err := fsutils.MkdirAll(i.cfg.Fs, constants.UkiCdromSource, os.ModeDir|os.ModePerm); err != nil {
```
Itxaka, line 41: no, this is wrong, if that path doesnt exists it means that immucore didnt mount the cdrom under there, so the next step wont work. We NEED that path mounted to be able to mount the following image, creating the dir here only hides the issue.

PR #227, opened 2024-02-09
fix: do not try to create a dir

mudler:
this should have been created by immucore, better failing

### DIFF
--- a/pkg/uki/install.go
+++ b/pkg/uki/install.go
@@ -1,7 +1,6 @@
 package uki
 
 import (
-	"errors"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/uki/install.go
```diff
@@ -36,13 +35,8 @@ func (i *InstallAction) Run() (err error) {
...
 	_, err = i.cfg.Fs.Stat(constants.UkiCdromSource)
-	if errors.Is(err, os.ErrNotExist) {
-		// The directory is under /run, so we can create it if not present
-		if err := fsutils.MkdirAll(i.cfg.Fs, constants.UkiCdromSource, os.ModeDir|os.ModePerm); err != nil {
-			i.cfg.Logger.Errorf("Error creating %s: %s", constants.UkiCdromSource, err)
-		}
-	} else if err != nil {
-		i.cfg.Logger.Errorf("stat error for '%s'. error: %s", constants.UkiCdromSource, err)
+	if err != nil {
+		i.cfg.Logger.Errorf("Mountpoint not ready '%s'. error: %s", constants.UkiCdromSource, err)
```
mudler, line 39:
```suggestion
		i.cfg.Logger.Errorf("mountpoint '%s' not ready. error: %s", constants.UkiCdromSource, err)
```

Itxaka (approved):
can you add an error log from mountpoint not being a mountpoint in line 59? we also need to return there if the `UkiCdromSource` is not mounted

PR #228, opened 2024-02-10
Adapt uki install to new rootfsbase

Itxaka:
If we know immucore is mounting the livecd under rootfsbase like dracut does, we can do size calculation and prepare the spec much earlier in the process

Part of: https://github.com/kairos-io/kairos/issues/2222

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -565,8 +565,27 @@ func ReadUkiResetSpecFromConfig(c *Config) (*v1.ResetUkiSpec, error) {
 }
 
 func NewUkiInstallSpec(cfg *Config) (*v1.InstallUkiSpec, error) {
+	var activeImg v1.Image
+
+	isoRootExists, _ := fsutils.Exists(cfg.Fs, constants.IsoBaseTree)
+
+	// First try to use the install media source
+	if isoRootExists {
+		activeImg.Source = v1.NewDirSrc(constants.IsoBaseTree)
+	}
+	// Then any user provided source
+	if cfg.Install.Source != "" {
+		activeImg.Source, _ = v1.NewSrcFromURI(cfg.Install.Source)
+	}
+	// If we dont have any just an empty source so the sanitation fails
+	// TODO: Should we directly fail here if we got no source instead of waiting for the Sanitize() to fail?
+	if !isoRootExists && cfg.Install.Source == "" {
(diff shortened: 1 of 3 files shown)


jimmykarily:
> If we know immucore is mounting the livecd under rootfsbase like dracut does

I don't see any changes related to this in this PR. Did this happen in another PR (the mounting)?

jimmykarily:
> > If we know immucore is mounting the livecd under rootfsbase like dracut does
> 
> I don't see any changes related to this in this PR. Did this happen in another PR (the mounting)?

Never mind, found it: https://github.com/kairos-io/immucore/pull/224

Itxaka:
> > > If we know immucore is mounting the livecd under rootfsbase like dracut does
> > 
> > 
> > I don't see any changes related to this in this PR. Did this happen in another PR (the mounting)?
> 
> Never mind, found it: [kairos-io/immucore#224](https://github.com/kairos-io/immucore/pull/224)

yep, sorry I updated the card at https://github.com/kairos-io/kairos/issues/2222 but didnt link this to it directly

Itxaka:
Manually tested and it works as expected.

pkg/config/spec.go
```diff
@@ -565,8 +565,27 @@ func ReadUkiResetSpecFromConfig(c *Config) (*v1.ResetUkiSpec, error) {
...
+	}
+	// Then any user provided source
+	if cfg.Install.Source != "" {
+		activeImg.Source, _ = v1.NewSrcFromURI(cfg.Install.Source)
+	}
+	// If we dont have any just an empty source so the sanitation fails
+	// TODO: Should we directly fail here if we got no source instead of waiting for the Sanitize() to fail?
+	if !isoRootExists && cfg.Install.Source == "" {
+		activeImg.Source = v1.NewEmptySrc()
+	}
```
mauromorales, line 584: are these assignments stepping on top of each other? i.e. are they in order of relevance (reverted)?

Itxaka: yes, this is a copy of the non-uki install spec generation. First the media device as a default, then if there is a explicit source from flags or config we override.

And if there is none, then we generate an empty source and let the sanitize fail due to that.

The empty stuff should probably be fixed, but it needs to be fixed on both installs as it changes the current workflow (let sanitize fail)

mauromorales (approved):

PR #230, opened 2024-02-14
Add shim to choose next entry to boot from

Itxaka:
Provide a shim to select the next entry for the system for the next boot.

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/kairos-io/kairos-agent/v2/pkg/constants"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -35,15 +36,6 @@ import (
 	"github.com/urfave/cli/v2"
 	"gopkg.in/yaml.v3"
 )
-
-var configScanDir = []string{
-	"/oem",
-	"/system/oem",
-	"/usr/local/cloud-config",
-	"/run/initramfs/live",
-	"/etc/kairos",    // Default system configuration file https://github.com/kairos-io/kairos/issues/2221
-	"/etc/elemental", // for backwards compatibility
-}
 
 // ReleasesToOutput gets a semver.Collection and outputs it in the given format
 // Only used here.
@@ -186,7 +178,7 @@ See https://kairos.io/docs/upgrade/manual/ for documentation.
 			}
 
 			return agent.Upgrade(source, c.Bool("force"),
-				c.Bool("strict-validation"), configScanDir,
+				c.Bool("strict-validation"), constants.GetConfigScanDirs(),
 				c.Bool("pre"), c.Bool("recovery"),
 			)
 		},
@@ -330,7 +322,7 @@ E.g. kairos-agent install-bundle container:quay.io/kairos/kairos...
 				Description: "Show the runtime configuration of the machine. It will scan the machine for all the configuration and will return the config file processed and found.",
 				Aliases:     []string{},
 				Action: func(c *cli.Context) error {
-					config, err := agentConfig.Scan(collector.Directories(configScanDir...), collector.NoLogs)
+					config, err := agentConfig.Scan(collector.Directories(constants.GetConfigScanDirs()...), collector.NoLogs)
 					if err != nil {
 						return err
 					}
@@ -359,7 +351,7 @@ enabled: true`,
 				Description: "It allows to navigate the YAML config file by searching with 'yq' style keywords as `config get k3s` to retrieve the k3s config block",
 				Aliases:     []string{"g"},
 				Action: func(c *cli.Context) error {
-					config, err := agentConfig.Scan(collector.Directories(configScanDir...), collector.NoLogs, collector.StrictValidation(c.Bool("strict-validation")))
+					config, err := agentConfig.Scan(collector.Directories(constants.GetConfigScanDirs()...), collector.NoLogs, collector.StrictValidation(c.Bool("strict-validation")))
 					if err != nil {
 						return err
 					}
@@ -430,7 +422,7 @@ enabled: true`,
 		},
 		Action: func(c *cli.Context) error {
 
-			config, err := agentConfig.Scan(collector.Directories(configScanDir...), collector.NoLogs, collector.StrictValidation(c.Bool("strict-validation")))
+			config, err := agentConfig.Scan(collector.Directories(constants.GetConfigScanDirs()...), collector.NoLogs, collector.StrictValidation(c.Bool("strict-validation")))
 			if err != nil {
 				return err
 			}
@@ -540,7 +532,7 @@ This command is meant to be used from the boot GRUB menu, but can be started man
 		Action: func(c *cli.Context) error {
 			source := c.String("source")
 
-			return agent.Install(source, configScanDir...)
+			return agent.Install(source, constants.GetConfigScanDirs()...)
 		},
 	},
 	{
@@ -584,7 +576,7 @@ This command is meant to be used from the boot GRUB menu, but can likely be used
 			unattended := c.Bool("unattended")
 			resetOem := c.Bool("reset-oem")
 
-			return agent.Reset(reboot, unattended, resetOem, configScanDir...)
+			return agent.Reset(reboot, unattended, resetOem, constants.GetConfigScanDirs()...)
 		},
 		Usage: "Starts kairos reset mode",
 		Description: `
@@ -663,7 +655,7 @@ The validate command expects a configuration file as its only argument. Local fi
 		},
 		Action: func(c *cli.Context) error {
 			stage := c.Args().First()
-			config, err := agentConfig.Scan(collector.Directories(configScanDir...), collector.NoLogs)
+			config, err := agentConfig.Scan(collector.Directories(constants.GetConfigScanDirs()...), collector.NoLogs)
 			config.Strict = c.Bool("strict")
 
 			if len(c.StringSlice("cloud-init-paths")) > 0 {
@@ -706,7 +698,7 @@ The validate command expects a configuration file as its only argument. Local fi
 			if err != nil {
 				return fmt.Errorf("invalid path %s", destination)
 			}
-			config, err := agentConfig.Scan(collector.Directories(configScanDir...), collector.NoLogs)
+			config, err := agentConfig.Scan(collector.Directories(constants.GetConfigScanDirs()...), collector.NoLogs)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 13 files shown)


main.go
```diff
@@ -745,6 +737,36 @@ The validate command expects a configuration file as its only argument. Local fi
...
+		},
+		Action: func(c *cli.Context) error {
+			cfg, err := agentConfig.Scan(collector.Directories(constants.GetConfigScanDirs()...), collector.NoLogs)
+			if err != nil {
+				return err
+			}
+			s := c.String("select")
+			// Here we should separate the call between uki and non-uki
+			// uki deals with systemd-boot and non-uki with grub
+			// so its a different path
```
mauromorales, line 762: should we maybe err out in such case?

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+	loaderConf, err := utils.SystemdBootConfReader(filepath.Join(efiPartition.MountPoint, "loader/loader.conf"))
+	if err != nil {
+		return err
+	}
+
+	entries, err := ListSystemdEntries(cfg, efiPartition)
+
+	// create a selector
+	selector := selection.New(fmt.Sprintf("Select Boot Entry (current entry: %s)", loaderConf["default"]), entries)
+	selector.Filter = nil        // Remove the filter
```
mauromorales, line 113: I think it would be better to have a `selector.RemoveFilter()` function so we don't need to rely on comments so much

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+	selector.ResultTemplate = `` // Do not print the result as we are asking for confirmation afterwards
```
mauromorales, line 114: something similar here

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+	for scanner.Scan() {
+		line := scanner.Text()
+		parts := strings.SplitN(line, " ", 2)
+		if len(parts) == 2 {
+			result[parts[0]] = parts[1]
+		}
+		if len(parts) == 1 {
+			result[parts[0]] = ""
+		}
+	}
```
mauromorales, line 146: if I remember correctly on the other story we were pairing with @jimmykarily we were using something similar, maybe it's already time to introduce this as an abstraction for both codes to share

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+		_ = file.Close()
+	}(file)
+
+	writer := bufio.NewWriter(file)
+	for k, v := range conf {
+		if v == "" {
+			_, err = writer.WriteString(fmt.Sprintf("%s \n", k))
+		} else {
+			_, err = writer.WriteString(fmt.Sprintf("%s %s\n", k, v))
+		}
```
mauromorales, line 171: if `v == ""` how come we need the two cases? I think it would always print `key \n`

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+			return err
+		}
+	}
+
+	return writer.Flush()
+}
+
+func ListSystemdEntries(cfg *config.Config, efiPartition *v1.Partition) ([]string, error) {
+	var entries []string
+	err := fsutils.WalkDirFs(cfg.Fs, filepath.Join(efiPartition.MountPoint, "loader/entries/"), func(path string, info os.DirEntry, err error) error {
```
mauromorales, line 182: This function is doing what I mentioned on the comment about the reader and writer functions. I think this one makes more sense because the logic of where those files exist is encapsulated

pkg/action/bootentries.go
Itxaka: Good point. This was mostly because I wanted to test if that is a valid conf and error out if otherwise

pkg/action/bootentries.go
Itxaka: Yes, this was originally here I think and moved into the ask but not merged yet and I didn't want to block this or force Dimitri's patch to get it, so this is here as a workaround to be able to test this and see the PR, but this functions belong in ~~a museum~~ the sdk

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+		cfg.Logger.Debugf("Checking file %s", path)
+		if info.IsDir() {
+			return nil
+		}
+		if filepath.Ext(info.Name()) != ".conf" {
+			return nil
+		}
+		entries = append(entries, info.Name())
+		return err
```
mauromorales, line 191: not making a comment to change, just to say that I find this a bit weird, since err is never set. Not sure if I prefer to have nil here to be clear that there's no chance of error, or if the variable needs to be used because of the compiler hehe

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+	// Remount it RO when finished
+	defer func(source string, target string, fstype string, flags uintptr, data string) {
+		err = syscall.Mount(source, target, fstype, flags, data)
+		if err != nil {
+			cfg.Logger.Errorf("could not remount EFI partition as RO: %s", err)
+		}
+	}("", efiPartition.MountPoint, "", syscall.MS_REMOUNT|syscall.MS_RDONLY, "")
+
+	systemdConf, err := SystemdBootConfReader(filepath.Join(efiPartition.MountPoint, "loader/loader.conf"))
```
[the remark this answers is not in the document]
Itxaka: Indeed but that would prevent those functions to be more generic in order to be used for other conf file that may not end up in the same place, i.e. the individual entries can be parsed by this function as well, as well as other configs in different places

So imo the file location as a parameter is good enough to be reused for the entries later

pkg/action/bootentries.go
```diff
@@ -0,0 +1,212 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+)
+
+func SelectBootEntry(cfg *config.Config, entry string) error {
+	// Read the systemd-boot conf file
+	if !strings.HasSuffix(entry, ".conf") {
+		entry = entry + ".conf"
```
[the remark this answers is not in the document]
Itxaka: This is in case the user selects an entry without the conf by using the --select flag, for example by calling it with active or passive.

Nicer UX imho.

main.go
Itxaka: I plan on expanding this to deal with grub systems as well, it's just not implemented yet :)

Itxaka: Sorry, got send before finishing. I don't plan to merge this yet

pkg/action/bootentries.go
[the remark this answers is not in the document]
mauromorales: i'm not sure, seems like these functions are very systemd specific, even the name suggests this, so I think it's better to encapsulate the logic within. See the comment on ListSystemdEntries. There could definitely be a higher level BootConfReader function that calls this one if it's on a systemd-boot system or a different one with GRUB

pkg/action/bootentries.go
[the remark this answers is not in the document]
mauromorales: gotcha, thanks for the explanation ... how about we make a function comesFromFlag() or something like that, or a comment is also good

main.go
mauromorales: gotcha, no worries

pkg/action/bootentries.go
mauromorales: good point, the refactoring can come later. I created a ticket https://github.com/kairos-io/kairos/issues/2252

pkg/action/bootentries.go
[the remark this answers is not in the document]
Itxaka: Ah yeah, that it's a good point. Let's see if I can add the grub part tomorrow and refactor this a bit. Plus tests!

pkg/action/bootentries.go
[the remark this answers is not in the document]
Itxaka: I dont think we can know if it comes from a flag or not, not I think its the function work to know if that comes from whatever source, but this is mainly sanitation just-in-case.

I'll add a comment there.

pkg/action/bootentries.go
Itxaka: nice, good catch xD

This should fail then? Where is this err coming from? I guess its currently the same as nil, but better make it explicit

pkg/constants/constants.go
```diff
@@ -149,3 +149,14 @@ func GetGrubFonts() []string {
 func GetGrubModules() []string {
 	return []string{"loopback.mod", "squash4.mod", "xzio.mod", "gzio.mod", "regexp.mod"}
 }
+
+func GetConfigScanDirs() []string {
```
Itxaka, line 153: extracted this from main into constants as it makes more sense here. Would probably make more sense to move it to sdk so all get that same dirs..

pkg/uki/reset.go
```diff
@@ -55,6 +56,13 @@ func (r *ResetAction) Run() (err error) {
 		}
 	}
 
+	// SelectBootEntry sets the default boot entry to the selected entry
+	err = action.SelectBootEntry(r.cfg, "active")
```
Itxaka, line 60: This requires the changes to the upgrade/install/reset workflow that @jimmykarily is doing, otherwise this will fail to set the active entry.

But the idea is that after reset is called, the next boot entry should be active always by default. Grub already does this, so systemd-boot needs to follow up the same behavior.

pkg/uki/reset.go
```diff
@@ -55,6 +56,13 @@ func (r *ResetAction) Run() (err error) {
...
+	// Should we fail? Or warn?
```
Itxaka, line 61: no idea here, maybe we should only warn and not fail? After all yes, the rest is done at this point and this is a minor thing....

pkg/utils/common.go
```diff
@@ -510,6 +510,17 @@ func IsUki() bool {
 	return false
 }
 
+// IsUkiWithFs checks if the system is running in UKI mode
+// by checking the kernel command line for the rd.immucore.uki flag
+// Uses a v1.Fs interface to allow for testing
+func IsUkiWithFs(fs v1.FS) bool {
```
Itxaka, line 516: this was needed in order to be able to abstract this check and be able to test it and actions that use it

pkg/uki/upgrade.go
```diff
@@ -77,6 +78,13 @@ func (i *UpgradeAction) Run() (err error) {
 		return err
 	}
 
+	// SelectBootEntry sets the default boot entry to the selected entry
+	err = action.SelectBootEntry(i.cfg, "active")
```
Itxaka, line 82: same as in reset

Itxaka:
Blocked by the reset/upgrade/install workflow changes to make it work as active/passive/recovery.

One that its merged this can be rebased to adjust to those changes.

pkg/utils/common.go
```diff
@@ -530,12 +541,14 @@ func UkiBootMode() state.Boot {
 }
 
 // SystemdBootConfReader reads a systemd-boot conf file and returns a map with the key/value pairs
-func SystemdBootConfReader(filePath string) (map[string]string, error) {
-	file, err := os.Open(filePath)
+func SystemdBootConfReader(fs v1.FS, filePath string) (map[string]string, error) {
```
mauromorales, line 544: there's a version on the sdk now https://github.com/kairos-io/kairos-sdk/commit/f666b183705b7aee442c287ca07fd8cd469d38bc#diff-0df82f8689d3e8d881be0653ab4e9f8221c5835ec95a14930d11669e9c7f07efR358

pkg/uki/reset.go
Itxaka: changes are in, this comment is now outdated!

mudler:
Also, let's not forget about docs :)

pkg/utils/common.go
Itxaka: problem with this is, the sdk version doesnt have an abstrac FS to execute the operations on, so that makes it impossible to test it correctly without touching the real OS dir underneath (i.e. on test, create the proper files so we can test that its reading them and so on)

Which means we lose testing on this part if we use that implementation.

Would it be ok to maintain this for 3.0.0 and then try to move at least the v1.FS and its utils to the sdk and reuse it on the functions there? So everything can use that abstract FS which makes it easier to test?

@jimmykarily @mauromorales @mudler

Itxaka:
docs: https://github.com/kairos-io/kairos-docs/pull/149

PR #231, opened 2024-02-20
Implement uki "kairos-agent upgrade --recovery"

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2263

I copy the recovery artifacts a bit more selectively to avoid overwriting any other files (e.g. the `loader.conf`)

### DIFF
--- a/pkg/uki/upgrade.go
+++ b/pkg/uki/upgrade.go
@@ -47,7 +47,7 @@ func (i *UpgradeAction) Run() (err error) {
 	// files, thus we take a simpler approach and only install the new efi file
 	// and the relevant conf
 	if i.spec.RecoveryUpgrade {
-		return i.installRecovery(i.cfg.Logger)
+		return i.installRecovery()
 	}
 
 	// Dump artifact to efi dir
@@ -85,7 +85,7 @@ func (i *UpgradeAction) Run() (err error) {
 
 // installRecovery replaces the "recovery" role efi and conf files with
 // the UnassignedArtifactRole efi and loader files from dir
-func (i *UpgradeAction) installRecovery(logger v1.Logger) error {
+func (i *UpgradeAction) installRecovery() error {
 	tmpDir, err := os.MkdirTemp("", "")
 	if err != nil {
 		return fmt.Errorf("creating a tmp dir: %w", err)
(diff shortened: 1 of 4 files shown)


pkg/uki/upgrade.go
```diff
@@ -74,3 +82,37 @@ func (i *UpgradeAction) Run() (err error) {
 
 	return nil
 }
+
+// installRecovery replaces the "recovery" role efi and conf files with
+// the UnassignedArtifactRole efi and loader files from dir
+func (i *UpgradeAction) installRecovery(logger v1.Logger) error {
```
Itxaka, line 88: Why passing a logger when there is one at i.cfg.logger?

pkg/uki/upgrade.go
```diff
@@ -74,3 +82,37 @@ func (i *UpgradeAction) Run() (err error) {
...
+
+	// Dump artifact to tmp dir
+	e := elemental.NewElemental(i.cfg)
+	_, err = e.DumpSource(tmpDir, i.spec.Active.Source)
+	if err != nil {
+		return err
+	}
+
+	err = copyFile(
+		filepath.Join(tmpDir, "EFI", "kairos", UnassignedArtifactRole+".efi"),
```
Itxaka, line 103: Maybe it's the review view, but where does this magic unassigned artifact role var comes from? I can't find it in this code? Is it a constant? Should it be?

mauromorales: IIRC it is defined in the package as a constant

mauromorales (approved):
lgtm

pkg/uki/upgrade.go
jimmykarily: https://github.com/kairos-io/kairos-agent/blob/58c039194f19de476cc0630a1b9350c85185e6c5/pkg/uki/common.go#L15

pkg/uki/upgrade.go
jimmykarily: oops good catch! Actually that's the logger I'm passing to the method. It was like that because initially I didn't make this a method of UpgradeAction (it had no receiver). I then moved it under `UpgradeAction` but didnt' notice I no longer needed to pass the logger. Fixing.

PR #236, opened 2024-02-27
Use new logger facilities

Itxaka:
Using the new shared logger from sdk to provide:
 - Same log format across all our apps
 - logging to /run and /var/log
 - Reduce complexity of having to manage the logger on each place

This PR also adds a hook after install to try to copy the install logs into the installed system persistent partition so they are accessible on the next reboots.


main.go
```diff
@@ -662,7 +659,7 @@ The validate command expects a configuration file as its only argument. Local fi
 				config.CloudInitPaths = append(config.CloudInitPaths, c.StringSlice("cloud-init-paths")...)
 			}
 			if c.Bool("debug") {
-				config.Logger.SetLevel(logrus.DebugLevel)
+				config.Logger.SetLevel("debug")
```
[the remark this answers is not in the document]
Itxaka: yes, but this is something that the old library didnt allow easily to do as was really weird.

Plus the idea would be to add an extra flag called log-level (keep debug as a shortcut) to start having normal log levels like normal people lol. So setting the level via string seems ok to me.

IsDebug no idea why is there, there is some code using it that should not use it I think. We should not care about the log level, we just send things to the log and thats it.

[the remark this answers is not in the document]
Itxaka: we could do something here but then it falls into us to do again the switch with any values we want, or use constants from other packages, I dunno. I like the string debug stuff, its understandable and easy to use and parse but I dont care enough, Im open to change it no problems.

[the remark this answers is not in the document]
mauromorales: nope I agree, we should just remove IsDebug and not try to have different paths depending on the log level

mauromorales (approved):

internal/agent/hooks/gruboptions.go
```diff
@@ -11,6 +11,9 @@ import (
 type GrubOptions struct{}
 
 func (b GrubOptions) Run(c config.Config, _ v1.Spec) error {
+	if len(c.Install.GrubOptions) == 0 {
```
Itxaka, line 14: this just skips runnig this if we dont have any options, makes no sense to run on empty options

internal/agent/hooks/logs.go
```diff
@@ -0,0 +1,47 @@
...
+	"path/filepath"
+	"syscall"
+)
+
+// CopyLogs copies all current logs to the persistent partition
+// useful during install to keep the livecd logs
+// best effort, no error handling
+type CopyLogs struct{}
+
+func (k CopyLogs) Run(c config.Config, _ v1.Spec) error {
```
Itxaka, line 19: This tries to mount persistent and store the logs into int when the install is finished so we can keep the logs after install :D

Done on best effort, if it cant do something it just fails silently with no issues. we dont want this breaking an install. Its nice to have but if its impossible for some reason it will just be like "ok" and continue with its mundane life as a go binary

internal/agent/hooks/kcrypt_uki.go
```diff
@@ -143,5 +149,32 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
 	if err != nil {
 		return err
 	}
+
+	// Copy logs to persistent partition
```
Itxaka, line 153: this makes more sense to me, we just open them once as we need to copy back the oem config and we take advantage of that time slot to copy the logs over if possible.
This reduces the lines of code needed for this, the lines of code that play with mounting unmounting and unlocking and the time the partitions are unlocked

internal/agent/hooks/kcrypt_uki.go
```diff
@@ -143,5 +149,32 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
...
+	c.Logger.Debug("Copying logs to persistent partition")
+	err = machine.Mount(constants.PersistentLabel, constants.PersistentDir)
+	if err != nil {
+		c.Logger.Errorf("could not mount persistent partition: %s", err)
+		return nil
+	}
+
+	// Create the directory on persistent if it doesn't exist
+	exists, _ := fsutils.Exists(c.Fs, filepath.Join(constants.PersistentDir, ".state", "var-log.bind"))
```
mauromorales, line 162: I think we don't need ot check if it exists

https://github.com/kairos-io/kairos-agent/blob/d139550b545d8d200094b2c57cf57a3d050fbe32/pkg/utils/fs/fs.go#L75

internal/agent/hooks/kcrypt_uki.go
```diff
@@ -143,5 +149,32 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
...
+	if !exists {
+		_ = fsutils.MkdirAll(c.Fs, filepath.Join(constants.PersistentDir, ".state", "var-log.bind"), 0755)
+	}
+	// Copy all current logs to the persistent partition
+	err = internalutils.SyncData(c.Logger, c.Runner, c.Fs, "/var/log/", filepath.Join(constants.PersistentDir, ".state", "var-log.bind"), []string{}...)
```
mauromorales, line 167: the joined path is used to create and here so we could put into a variable

internal/agent/hooks/kcrypt_uki.go
Itxaka: good one

mauromorales (approved):

PR #240, opened 2024-03-01
Bump to use vfs4

Itxaka:
Requires the latest yip version as that one has been bumped to vfs5 as well


mauromorales (approved):

pkg/utils/chroot.go
```diff
@@ -149,12 +148,11 @@ func (c *Chroot) RunCallback(callback func() error) (err error) {
 	}()
 
 	// Store current root
-	oldRootF, err = c.config.Fs.Open("/")
+	rootPath, _ := c.config.Fs.RawPath("/")
```
Itxaka, line 151: we no longer need to open the oldRoot directly to be able to chdir to it later, which is nice

pkg/utils/common.go
```diff
@@ -163,6 +163,14 @@ func SyncData(log sdkTypes.KairosLogger, runner v1.Runner, fs v1.FS, source stri
 		}
 		if t, err := fs.RawPath(target); err == nil {
 			target = t
+			// create target path if it doesnt exists
+			if _, err := os.Stat(target); err != nil {
```
Itxaka, line 167: looks like older vfs was more strict when dir were not there. Now its more relaxed and it doesnt fail so we need to check ourselves.

pkg/utils/common.go
```diff
@@ -451,7 +463,8 @@ func FindFileWithPrefix(fs v1.FS, path string, prefixes ...string) (string, erro
 		}
 		for _, p := range prefixes {
 			if strings.HasPrefix(f.Name(), p) {
-				if f.Mode()&os.ModeSymlink == os.ModeSymlink {
+				info, _ := f.Info()
+				if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
Itxaka, line 467: now we dont have the mode directly so we need to info it

pkg/utils/fs/fs.go
```diff
@@ -236,12 +236,7 @@ func walkDir(fs v1.FS, path string, d fs.DirEntry, walkDirFn fs.WalkDirFunc) err
 }
 
 func readDir(fs v1.FS, dirname string) ([]fs.DirEntry, error) {
-	f, err := fs.Open(dirname)
-	if err != nil {
-		return nil, err
-	}
-	dirs, err := f.ReadDir(-1)
-	f.Close()
+	dirs, err := fs.ReadDir(dirname)
```
Itxaka, line 239: now it directly opens the dir, this function is mainly obsolete as it does the same as the underlying!

mudler (approved):
looking good here!

PR #241, opened 2024-03-06
Retry unlock if we cant find the unlocked partitions

Itxaka:
This si something that happens rarely but could happen. If we unlockAll but at that time the kernel partition table is not refreshed it could happen that we only unlock one partition, thus retrying to find the unlocked one will never succeed. Instead we should try unlocking on each iteration to see if they appear.


Itxaka:
Needs testing. It should not fail the hook but I dont know what happens if partitions are unlocked already and we try to unlock them again, I just hope they dont get duplicated or something.

internal/agent/hooks/kcrypt_uki.go
```diff
@@ -130,6 +130,10 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
 			if part == "" {
 				c.Logger.Infof("Partition %s not found, waiting %d seconds before retrying", p, i)
 				time.Sleep(time.Duration(i) * time.Second)
+				// Retry the unlock as well, because maybe the partition was not refreshed on time for unlock to unlock it
+				// So no matter how many tries we do, it will still be locked and will never appear
+				err := kcrypt.UnlockAll(true)
+				c.Logger.Debugf("UnlockAll returned: %s", err)
```
mauromorales, line 136: should we maybe add this Debug message also in the first UnlockAll, to see if the initial one already locked everything

mauromorales (approved):

internal/agent/hooks/kcrypt_uki.go
Itxaka: this is actually wrong, we should check if the err is not nil before printing absurd empty stuff lol

This line was autosuggested by copilot, thats what you get for not reviewing stuff arg

jimmykarily (approved):

PR #244, opened 2024-03-07
Make sure kcrypt hook umounts persistent

Itxaka:
We cant know its status so we better make sure its unmounted.

Also adds some logging around the hooks to see what they run.

Also adds back the CopyLogs hook that was lost in a rebase probably.

### DIFF
--- a/internal/agent/hooks/logs.go
+++ b/internal/agent/hooks/logs.go
@@ -1,13 +1,17 @@
 package hook
 
 import (
+	"fmt"
 	"github.com/kairos-io/kairos-agent/v2/pkg/config"
 	"github.com/kairos-io/kairos-agent/v2/pkg/constants"
 	v1 "github.com/kairos-io/kairos-agent/v2/pkg/types/v1"
 	internalutils "github.com/kairos-io/kairos-agent/v2/pkg/utils"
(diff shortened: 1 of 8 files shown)


mauromorales (approved):

internal/agent/hooks/logs.go
```diff
@@ -17,9 +21,31 @@ import (
 type CopyLogs struct{}
 
 func (k CopyLogs) Run(c config.Config, _ v1.Spec) error {
+	c.Logger.Logger.Debug().Msg("Running CopyLogs hook")
 	c.Logger.Debugf("Copying logs to persistent partition")
 	_ = machine.Umount(constants.PersistentDir)
 
+	// Path if we have encrypted persistent
+	if len(c.Install.Encrypt) != 0 {
+		err := kcrypt.UnlockAll(false)
```
jimmykarily, line 30: We are in `CopyLogs` and we are decrypting partitions? This feels wrong...

Itxaka: yes, if you have encrypted partitions you need to decrypt them to be able to work on them, otherwise you cant copy the logs on them.

If our mount would auto do this, it would be nice :D

jimmykarily: gotcha

PR #245, opened 2024-03-11
Run after reset hook when resetting in UKI mode

jimmykarily:
To allow the workaround described here:
https://github.com/kairos-io/kairos/issues/2304

### DIFF
--- a/pkg/uki/reset.go
+++ b/pkg/uki/reset.go
@@ -78,6 +78,18 @@ func (r *ResetAction) Run() (err error) {
 		return err
 	}
 
+	if mnt, err := elementalUtils.IsMounted(r.cfg, r.spec.Partitions.OEM); !mnt && err == nil {
+		err = e.MountPartition(r.spec.Partitions.OEM)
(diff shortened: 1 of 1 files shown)


pkg/uki/reset.go
```diff
@@ -78,6 +78,18 @@ func (r *ResetAction) Run() (err error) {
 		return err
 	}
 
+	if mnt, err := elementalUtils.IsMounted(r.cfg, r.spec.Partitions.OEM); !mnt && err == nil {
+		err = e.MountPartition(r.spec.Partitions.OEM)
```
jimmykarily, line 82: Otherwise the config is not being read

jimmykarily:
I created a test that should now be red and should become green with this fix: https://github.com/kairos-io/kairos/pull/2331

I still need to verify locally.

Itxaka (approved):

mudler (approved):

mudler:
I guess we can merge this @jimmykarily ?

PR #247, opened 2024-03-13
Use existing role names for bootentry command on UKI

mauromorales:
Switch from active.conf, passive.conf and recovery.conf to cos, fallback and recovery respectively

relates to kairos-io/kairos#2325


Itxaka (commented):
I wonder if this would work with custom cmdlines and such?

mauromorales:
@Itxaka very good point, it won't :P ... let me see if I can make the necessary changes

mauromorales:
extended the functionality to work with extended and extra cmdline, added tests for all the scenarios

jimmykarily:
These are some crazy naming requirements :D. In any case, good job @mauromorales !

jimmykarily (approved):

PR #250, opened 2024-03-15
Mount efi partition if not mounted for bootentries

Itxaka:
So we can run it on livecd after the install or in other places (reset)

Fixes: https://github.com/kairos-io/kairos/issues/2334

### DIFF
--- a/pkg/action/bootentries.go
+++ b/pkg/action/bootentries.go
@@ -241,7 +241,7 @@ func listBootEntriesSystemd(cfg *config.Config) error {
 		return err
 	}
 	// mount if not mounted
-	if mounted, _ := utils.IsMounted(cfg, efiPartition); !mounted {
+	if mounted, err := utils.IsMounted(cfg, efiPartition); !mounted && err == nil {
 		if efiPartition.MountPoint == "" {
 			efiPartition.MountPoint = cnst.EfiDir
 		}
--- a/pkg/action/bootentries.go
+++ b/pkg/action/bootentries.go
@@ -2,6 +2,8 @@ package action
 
 import (
 	"fmt"
+	cnst "github.com/kairos-io/kairos-agent/v2/pkg/constants"
+	"github.com/kairos-io/kairos-agent/v2/pkg/elemental"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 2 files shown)


mudler (approved):

pkg/action/bootentries.go
```diff
@@ -229,11 +231,28 @@ func bootNameToSystemdConf(name string) (string, error) {
...
+	e := elemental.NewElemental(cfg)
+	cleanup := utils.NewCleanStack()
+	defer func() { err = cleanup.Cleanup(err) }()
 	// Get EFI partition
 	efiPartition, err := partitions.GetEfiPartition()
 	if err != nil {
 		return err
 	}
+	// mount if not mounted
+	if mounted, _ := utils.IsMounted(cfg, efiPartition); !mounted {
```
jimmykarily, line 244: we should better not skip the error here. After all, we are checking for `false` which could be returned in case of an error, thus entering the block when things don't even work.

This methods already has error handling so let's handle this one too.

pkg/uki/install.go
```diff
@@ -159,6 +160,9 @@ func (i *InstallAction) Run() (err error) {
 		return fmt.Errorf("removing artifact set with role %s: %w", UnassignedArtifactRole, err)
 	}
 
+	// SelectBootEntry sets the default boot entry to the selected entry
+	err = action.SelectBootEntry(i.cfg, "active")
```
mauromorales, line 164: also error handling afterwards, no? or at least logging if it's not meant to break

pkg/action/bootentries.go
Itxaka: good point, done!

pkg/uki/install.go
Itxaka: good point, done

pkg/action/bootentries.go
```diff
@@ -229,11 +231,28 @@ func bootNameToSystemdConf(name string) (string, error) {
...
+		if efiPartition.MountPoint == "" {
+			efiPartition.MountPoint = cnst.EfiDir
+		}
+		err = e.MountPartition(efiPartition)
+		if err != nil {
+			cfg.Logger.Errorf("could not mount EFI partition: %s", err)
+			return err
+		}
+		cleanup.Push(func() error { return e.UnmountPartition(efiPartition) })
+	}
```
jimmykarily, line 254: log the errors if any?

Itxaka: That gets auto logged by the clean function IIRC?

Itxaka: but this is inside the clean stack, and that auto reports errors I think. This is done everywhere and IIRC always works at the end, it will report the errors while executing the clean function  stack

PR #251, opened 2024-03-16
After upgrade or recovery, set bootentry to "cos"

mauromorales:
Previously we used to call this active but since the bootentry change, this should be cos instead

fixes kairos-io/kairos#2354

### DIFF
--- a/pkg/action/bootentries.go
+++ b/pkg/action/bootentries.go
@@ -2,14 +2,15 @@ package action
 
 import (
 	"fmt"
-	cnst "github.com/kairos-io/kairos-agent/v2/pkg/constants"
-	"github.com/kairos-io/kairos-agent/v2/pkg/elemental"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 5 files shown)


mauromorales:
I haven't had the chance to test this yet, but I think will properly fix it

Itxaka (commented):
I think install also has that call.

Should we make it work with active as well? So if it's called with active we move it to cos?

Or drop cos fully and call it active?

mauromorales:
@Itxaka shit, I swear my vscode was not showing it on the search :shrug: 

i wouldn't drop cos yet because then we also have to change it on the GRUB side, but should definitly be a change for one of our next minor releases

mauromorales:
might not be a bad idea to add support for active already, let me have a look

Itxaka:
Ahh forgot about grub yikes

mauromorales:
@Itxaka ok added support to also select "active"

Itxaka (approved):

PR #281, opened 2024-04-08
Remove config show in favor of parent command

mauromorales:


jimmykarily:
I'm wondering if we should keep the subcommand with a deprecation warning for a couple of versions.

mauromorales:
@jimmykarily yup maybe that's the better approach, I'll change it. But we lack a deprecation process atm, so my guess is that, it will just stay with that message until who knows when. Whenever we add a deprecation warning, we should know that the deprecation is then planned for the upcoming N major/minor version 100% and have a ticket for it.

jimmykarily:
> @jimmykarily yup maybe that's the better approach, I'll change it. But we lack a deprecation process atm, so my guess is that, it will just stay with that message until who knows when. Whenever we add a deprecation warning, we should know that the deprecation is then planned for the upcoming N major/minor version 100% and have a ticket for it.

Yes we lack a general deprecation policy. For now, just name a version (current +2 minors?) in the deprecation notice. Even if we loose the ticket, anyone looking at the message (or the code) will be able to tell if it can be removed or not. I think we've done something similar with some cli flags in the past.

jimmykarily (approved):

PR #282, opened 2024-04-09
added null-logger to render-template command

Akvanvig:
Adds nulllogger created for config get command to render-template command as well

Fixes: https://github.com/kairos-io/kairos/issues/2449


Itxaka (approved):

Itxaka:
good one @Akvanvig !

PR #289, opened 2024-04-11
Fix url when manual install

Itxaka:
According to docs you can pass a file or an http file to the manual-install command, but we were not respecting that.

This patch makes it so a local file or an http file is properly working when doing a manual install.

If its a file, read it directly and pass it to the Scan as a reader

If its an http file, generate a config_url value and pass it to the reader

Fixes: https://github.com/kairos-io/kairos/issues/2424


internal/agent/install_test.go
```diff
@@ -13,56 +11,16 @@ import (
...
 	. "github.com/onsi/gomega"
+	"github.com/twpayne/go-vfs/v4/vfst"
 )
 
 const printOutput = `BYT;
 /dev/loop0:50593792s:loopback:512:512:gpt:Loopback device:;`
 const partTmpl = `
 %d:%ss:%ss:2048s:ext4::type=83;`
 
-var _ = Describe("prepareConfiguration", func() {
```
mauromorales, line 28: I think we should keep a test, even if this one somehow missed the original requirement, at least it makes it clear what we actually want to have working here

Itxaka: Got it, yeah. Not sure why I removed the function. I should just adapt it to return an io.reader

jimmykarily (approved):

internal/agent/install.go
```diff
@@ -54,25 +54,33 @@ func displayInfo(agentConfig *Config) {
...
+	// If its a local file, we jsut read it and pass it to the config.Scan
+	// If its a url, we need to create a configuration with the url and let the config.Scan handle it
+	if u, err := url.Parse(c); err != nil || u.Scheme == "" {
+		file, err := os.ReadFile(c)
+		if err != nil {
+			return err
+		}
+		cfg = bytes.NewReader(file)
+	} else {
+		cfgUrl := fmt.Sprintf(`config_url: %s`, c)
```
jimmykarily, line 70: I still don't see why the old code wasn't working. It was constructing a file on disk with the exact same option no?

mauromorales (approved):

internal/agent/install.go
```diff
@@ -324,32 +321,26 @@ func ensureDataSourceReady() {
...
-	// defer cleanup until after parent is done
-	go func() {
-		<-ctx.Done()
-		_ = os.RemoveAll(f.Name())
-	}()
-
-	cfg := config.Config{
-		ConfigURL: source,
-	}
-	if err = yaml.NewEncoder(f).Encode(cfg); err != nil {
```
Itxaka, line 348: may be this was the issue? We never closed the file as the comments in NewEncoder mention:

// NewEncoder returns a new encoder that writes to w.
// The Encoder should be closed after use to flush all data
// to w.


So maybe its was not fully written?

Itxaka: cant even find anything else related to this that could explain the issues, unless we have a bug in the collector in which the dir scan gets overwritten fully by the readers scan?

Itxaka: will check that just in case as well...

PR #307, opened 2024-04-23
Add missing hooks to after install uki

Itxaka:
we were missing the bundle stuff, mounts (for custom mounts) and lifecycle

### DIFF
--- a/internal/agent/hooks/hook.go
+++ b/internal/agent/hooks/hook.go
@@ -32,7 +32,11 @@ var FirstBoot = []Interface{
 }
 
 // AfterUkiInstall sets which Hooks to run after uki runs the install action
-var AfterUkiInstall = []Interface{}
+var AfterUkiInstall = []Interface{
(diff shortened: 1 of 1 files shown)


mauromorales (approved):
Good catch

internal/agent/hooks/hook.go
```diff
@@ -32,7 +32,11 @@ var FirstBoot = []Interface{
 }
 
 // AfterUkiInstall sets which Hooks to run after uki runs the install action
-var AfterUkiInstall = []Interface{}
+var AfterUkiInstall = []Interface{
+	&BundlePostInstall{},
+	&CustomMounts{},
+	&Lifecycle{},
```
mauromorales, line 38: What about `&CopyLogs{}` ?

Itxaka: In uki, the logs are copied as part of the encryption jhook as to not keeping closing and opening the encrypted devices several times.

PR #309, opened 2024-04-24
Centralize uki hooks for encryption reasons

Itxaka:
This makes the hooks work without modifications to them, as they already mount/unmount oem/persistent and do it while they are unlocked

### DIFF
--- a/internal/agent/hooks/kcrypt_uki.go
+++ b/internal/agent/hooks/kcrypt_uki.go
@@ -3,10 +3,8 @@ package hook
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/kairos-io/kairos-agent/v2/pkg/config"
@@ -21,7 +19,7 @@ import (
 
 type KcryptUKI struct{}
 
-func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
+func (k KcryptUKI) Run(c config.Config, spec v1.Spec) error {
 	// pre-check for systemd version, we need something higher or equal to 252
 	run, err := utils.SH("systemctl --version | head -1 | awk '{ print $2}'")
 	systemdVersion := strings.TrimSpace(string(run))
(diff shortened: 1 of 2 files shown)


internal/agent/hooks/kcrypt_uki.go
```diff
@@ -163,33 +161,16 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
...
-	// Create the directory on persistent
-	err = fsutils.MkdirAll(c.Fs, varLog, 0755)
-	if err != nil {
-		c.Logger.Errorf("could not create directory on persistent partition: %s", err)
-		return nil
-	}
-	// Copy all current logs to the persistent partition
-	err = internalutils.SyncData(c.Logger, c.Runner, c.Fs, "/var/log/", varLog, []string{}...)
+	c.Logger.Logger.Debug().Msg("Finish KcryptUKI hook")
+	// We now have the partitions unlocked and ready, lets call the other hooks here instead of closing and reopening them each time
```
mauromorales, line 165: so if I understand correctly, this will happen on every kcrypt run instead of just in the after install one

internal/agent/hooks/kcrypt_uki.go
```diff
@@ -163,33 +161,16 @@ func (k KcryptUKI) Run(c config.Config, _ v1.Spec) error {
...
+	err = BundlePostInstall{}.Run(c, spec)
 	if err != nil {
-		c.Logger.Errorf("could not copy logs to persistent partition: %s", err)
-		return nil
+		return err
 	}
-	err = machine.Umount(constants.PersistentDir)
+	_ = CustomMounts{}.Run(c, spec)
```
mauromorales, line 170: this is the same than the mount/umount done on the left side?

mauromorales: ah no, it was just needed to copy the logs if i understand correctly, this is the CustomMounts you wanted to add, and the CopyLogs is doing this mount/unmount and moving logs

mauromorales (approved):

internal/agent/hooks/kcrypt_uki.go
Itxaka: no, the kcrypt_uki hook is only run after the install so it will only run once, after the install is done!

internal/agent/hooks/kcrypt_uki.go
Itxaka: yes, the removed part is the Copylogs hook, which was directly used in the code. Now we just call the hook as it was exactly the smae code repeated

internal/agent/hooks/kcrypt_uki.go
mauromorales: i see, a bit confusing 😅 so it's not really after install but a bit before

Itxaka: Yes, we had to change it for the after I stall stage to have access to the encrypted partitions

PR #337, opened 2024-05-16
Add method to validate signature of efi file

Itxaka:
This seems rigth but needs testing. Also makes sense to have this in the sdk instead maybe?

Fixes https://github.com/kairos-io/kairos/issues/2200


pkg/uki/common.go
```diff
@@ -153,3 +159,79 @@ func copyFile(src, dst string) error {
...
+		logger.Error(fmt.Errorf("failed to read file %s: %s", fullArtifact, err))
+		return fmt.Errorf("failed to read file %s: %s", fullArtifact, err)
+	}
+	if !ok {
+		logger.Error("invalid pe header")
+		return fmt.Errorf("invalid pe header")
+	}
+
+	// We need to read the current db database to have the proper certs to check against
+	db, err := efi.Getdb()
```
jimmykarily, line 190: Ideally, we would need to check that the certificate signing the efi file is not blacklisted in dbx and that the binary signature is not blacklisted either. Even worse, we would need to check that the KEK and PK certs that sign the db that signs the efi file, are not blacklisted. This last one though, [I'm having trouble verifying](https://github.com/kairos-io/kairos/issues/2429).

Itxaka: good point, dbx should be checked to see if the cert is not in there.

Regarding the PK/KEK, can they be blacklisted but not the db? I mean, db+dbx is the way of checking the signatures in files no? Any other verification, needs to be done by the firmware itself to allow you updating db/dbx no?

In any case, doesnt look like sbctl is actually checking the dbx/pk/kek either, only checks if its signed with the db cert....

pkg/uki/common.go
```diff
@@ -153,3 +159,104 @@ func copyFile(src, dst string) error {
...
+	if err != nil {
+		logger.Logger.Error().Err(err).Msg("getting DB certs")
+		return err
+	}
+
+	x509CertDb, err := util.ReadCert(db.Bytes())
```
jimmykarily, line 208: same thing here (multiple certs possible) although we could pretend that it's a feature to only have the kairos cert in db (as in "more secure"). We can't tell the same for dbx though because one may even want to add old kairos digests or certs in dbx, so having multiple there is a good thing.

pkg/uki/common.go
```diff
@@ -153,3 +159,104 @@ func copyFile(src, dst string) error {
...
+			if ok {
+				// Signature is not on DBX but its on DB, verified
+				return nil
+			}
+		}
+	}
+	return errors.New("not ok")
+}
+
+func checkValidEfiHeader(r io.Reader) (bool, error) {
```
jimmykarily, line 248: Why not use it [from here](https://github.com/Foxboron/sbctl/blob/78476facea5e1c0bc9d1f47955b113d88c0a7928/util.go#L135)? It's exported.

pkg/uki/common_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Uki utils", Label("uki", "utils"), func() {
+	It("Check valid signature in file", func() {
+		checkArtifactsignatureIsValid()
+		Expect(true).To(BeTrue())
```
jimmykarily, line 11: Increasing our test coverage are we? :D

pkg/uki/common.go
```diff
@@ -153,3 +159,104 @@ func copyFile(src, dst string) error {
 
 	return destinationFile.Close()
 }
+
+// checkArtifactsignatureIsValid checks that a give efi artifact is signed properly with a signature that would allow it to
+// boot correctly in the current node if secureboot is enabled
+func checkArtifactsignatureIsValid(fs v1.FS, dir string, artifact string, logger sdkTypes.KairosLogger) error {
```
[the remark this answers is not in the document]
Itxaka: good point, will do

pkg/uki/common.go
```diff
@@ -153,3 +159,104 @@ func copyFile(src, dst string) error {
...
+	}
+
+	// First check if its on dbx, if it matches, then it means that its not valid and we can return early
+	dbx, err := efi.Getdbx()
+	if err != nil {
+		logger.Logger.Error().Err(err).Msg("getting DBx certs")
+		return err
+	}
+
+	x509CertDbx, err := util.ReadCert(dbx.Bytes())
```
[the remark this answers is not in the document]
Itxaka: yeah, I mainly did this looking at sbctl code, they do the same. No idea why, if they are just checking against their own cert, as they do look at the stored cert under /usr/share/secureboot so maybe we need to expand on this and get ALL the certs in the db/dbx :(

I just need to add tests to this with some signed files and some fake db/dbx vars in order to confirm this.

pkg/uki/common.go
Itxaka: because those 15 lines brings the whole sbctl as a dep and I dont want to add that extra dep for this few lines.

jimmykarily: Consider adding a comment that this method exists there. In case sbctl is imported in the future as a library for other reasons, we could then use it from there.

Itxaka: good point, will add!

Itxaka:
Seems to me like this has gottena  bit big. Just to parse a couple of things, we need to bring a library for parsing pe files, so for me it makes more sense to have this in the sdk as a generic function that checks the signature validity instead of cluttering teh agent with this maybe... ?

pkg/uki/common_test.go
```diff
@@ -0,0 +1,137 @@
...
+		file, err := os.ReadFile("tests/db-wrong")
+		Expect(err).ToNot(HaveOccurred())
+		err = fs.WriteFile(filepath.Join("/sys/firmware/efi/efivars", dbFile), file, os.ModePerm)
+		Expect(err).ToNot(HaveOccurred())
+		err = checkArtifactSignatureIsValid(fs, "/efitest.signed.efi", logger)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("not ok"))
+	})
+
+	It("Signatures does match in DB", func() {
```
jimmykarily, line 96: I kept reading "does not match" :D 
```suggestion
	It("Signature matches the DB", func() {
```

pkg/uki/common_test.go
```diff
@@ -0,0 +1,137 @@
...
+		dbFile := fmt.Sprintf("db-%s", attributes.EFI_IMAGE_SECURITY_DATABASE_GUID.Format())
+		file, err := os.ReadFile("tests/db")
+		Expect(err).ToNot(HaveOccurred())
+		err = fs.WriteFile(filepath.Join("/sys/firmware/efi/efivars", dbFile), file, os.ModePerm)
+		Expect(err).ToNot(HaveOccurred())
+		err = checkArtifactSignatureIsValid(fs, "/efitest.signed.efi", logger)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("Signatures does match in DB and not in DBX", func() {
```
jimmykarily, line 106:
```suggestion
	It("matches the DB and not DBX", func() {
```

Itxaka:
Seems to be working with a proper EFI:

```
2024-05-20T13:29:21Z INF Copying ttl.sh/kairosupgrade:latest source to /efi
2024-05-20T13:29:47Z INF Finished copying ttl.sh/kairosupgrade:latest into /efi
2024-05-20T13:29:47Z INF Checking artifact for valid signature what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z DBG Reading artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z DBG Parsing PE artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z DBG Checking if its an EFI file what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z DBG Getting DB certs what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z DBG Getting signatures from artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z DBG Getting DBX certs what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z DBG checking signature subject=ITXAKA what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:47Z INF verified subject=ITXAKA what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:29:48Z DBG Conf file /efi/loader/entries/passive.conf has values map[string]string{
  "efi": "/EFI/kairos/active.efi",
```

Itxaka:
certs in the machine DB:

```
root@localhost:~# mokutil --db
[key 1]
SHA1 Fingerprint: a3:d8:33:c9:02:af:6a:a2:65:34:71:14:d5:0c:a4:9c:62:d7:87:b2
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            0a:24:ee:5f:00:53:3e:94:1a:03:18:31:32:c8:2c:63:3e:0b:94:4b
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: CN=ITXAKA
        Validity
            Not Before: Apr 26 08:11:52 2024 GMT
            Not After : Apr 26 08:11:52 2025 GMT
        Subject: CN=ITXAKA
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                Public-Key: (2048 bit)
```

Itxaka:
Artifact checks out, signed by my default test key

```
root@localhost:~# sbverify /efi/EFI/kairos/active.efi --list
signature 1
image signature issuers:
 - /CN=ITXAKA
image signature certificates:
 - subject: /CN=ITXAKA
```

Itxaka:
Now testing with an upgrade image signed with a different key (generic one created on the fly during artifact build):

```
$ sbverify --list EFI/kairos/norole.efi 
signature 1
image signature issuers:
 - /CN=Test
image signature certificates:
 - subject: /CN=Test
   issuer:  /CN=Test
```

Check fails as expected:

```
2024-05-20T13:38:40Z INF Copying ttl.sh/kairosupgrade2:latest source to /efi
2024-05-20T13:39:07Z INF Finished copying ttl.sh/kairosupgrade2:latest into /efi
2024-05-20T13:39:07Z INF Checking artifact for valid signature what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG Reading artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG Parsing PE artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG Checking if its an EFI file what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG Getting DB certs what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG Getting signatures from artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG Getting DBX certs what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG checking signature subject=ITXAKA what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG checking signature subject="Microsoft Corporation UEFI CA 2011" what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z DBG checking signature subject="Microsoft Windows Production PCA 2011" what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:39:07Z ERR Checking signature error="not ok"
2024-05-20T13:39:07Z DBG Mounting partition COS_GRUB
1 error occurred:
	* not ok
```


Need to fix the error message though as its pretty crude LOL

Itxaka:
New error message a bit more extensive

```
2024-05-20T13:44:57Z INF Copying ttl.sh/kairosupgrade2:latest source to /efi
2024-05-20T13:45:29Z INF Finished copying ttl.sh/kairosupgrade2:latest into /efi
2024-05-20T13:45:29Z INF Checking artifact for valid signature what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:29Z DBG Reading artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:29Z DBG Parsing PE artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:29Z DBG Checking if its an EFI file what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:30Z DBG Getting DB certs what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:30Z DBG Getting signatures from artifact what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:30Z DBG Getting DBX certs what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:30Z DBG checking signature subject=ITXAKA what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:30Z DBG checking signature subject="Microsoft Corporation UEFI CA 2011" what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:30Z DBG checking signature subject="Microsoft Windows Production PCA 2011" what=/efi/EFI/Kairos/norole.efi
2024-05-20T13:45:30Z ERR Checking signature before upgrading error="could not find a signature in EFIVars DB that matches the upgrade artifact"
2024-05-20T13:45:30Z WRN Upgrade artifact signature does not match, upgrading to this source would result in an unbootable active system
Check the upgrade source and confirm that its signed with a valiod key, that key is in the machine DB and it has not been blacklisted.
```

pkg/uki/upgrade.go
```diff
@@ -64,6 +64,17 @@ func (i *UpgradeAction) Run() (err error) {
 		return err
 	}
 
+	// Check if the upgrade artifact contains the proper signature before copying
+	err = checkArtifactSignatureIsValid(i.cfg.Fs, filepath.Join(constants.UkiEfiDir, "EFI/Kairos", fmt.Sprintf("%s.efi", UnassignedArtifactRole)), i.cfg.Logger)
```
mauromorales, line 68: shouldn't this be? I guess it does not matter since it will never run on windows :P 

```suggestion
	err = checkArtifactSignatureIsValid(i.cfg.Fs, filepath.Join(constants.UkiEfiDir, "EFI", "Kairos", fmt.Sprintf("%s.efi", UnassignedArtifactRole)), i.cfg.Logger)
```

pkg/uki/upgrade.go
```diff
@@ -64,6 +64,17 @@ func (i *UpgradeAction) Run() (err error) {
...
+	if err != nil {
+		// Remove efi file to not occupy space and leave stuff around
+		_ = removeArtifactSetWithRole(i.cfg.Fs, constants.UkiEfiDir, UnassignedArtifactRole)
```
mauromorales, line 71: is there any value in loging the error instead of dismissing it?

pkg/uki/upgrade.go
```diff
@@ -64,6 +64,17 @@ func (i *UpgradeAction) Run() (err error) {
...
+		i.cfg.Logger.Logger.Error().Err(err).Msg("Checking signature before upgrading")
+		i.cfg.Logger.Logger.Warn().Msg("Upgrade artifact signature does not match, upgrading to this source would result in an unbootable active system.\n" +
```
mauromorales, line 73: would it be worth outputing signatures?

Itxaka: which one? The one in the artifact or the one in the checked databases?
And what part of the signature? Common name? Sha? its a bit weird as I have no idea what to print there

pkg/uki/upgrade.go
Itxaka: probably yes, will add

pkg/uki/common.go
```diff
@@ -153,3 +161,140 @@ func copyFile(src, dst string) error {
...
+
+	logger.Logger.Debug().Str("what", artifact).Msg("Getting DBX certs")
+	dbx, err := efi.Getdbx()
+	if err != nil {
+		logger.Logger.Error().Err(err).Msg("getting DBX certs")
+		return err
+	}
+
+	// First check the dbx database as it has precedence, on match, return immediately
+	for _, k := range *dbx {
```
mauromorales, line 250: this logic seems to me like it has to be similar to one we already have, can we somehow have a single function for this if that is the case?

pkg/uki/upgrade.go
mauromorales: @Itxaka yeah I see your point, thinking out loud here so bare with me lol ... ideally I'd like to see something that let's me distinguish between the two, like: wants and got, e.g.

```
want (key) sha -> got (key) sha
```

or something of sorts

Itxaka: on debug mode it will output the signature that its checking vs the one in the artifact, but it cna be very very noise, especially with the dbx database as it can have unlimited entries. 

Is that good enough?

pkg/uki/common.go
Itxaka: We got something like this in `dbCerts := signatures.ExtractCertsFromSignatureDatabase(db)` which gets the signatures but for DBX is not exactly similar as it can be of different types (hash, cert or even a third option which I have not encountered) so this is a bit more convoluted that the normal PK/DB/KEK so I would leave it as is as its very specific to this usecase.

pkg/uki/upgrade.go
mauromorales: yup sounds good, I just throwing ideas

mauromorales (approved):

PR #349, opened 2024-05-23
Bump sdk to v0.1.8

mauromorales:
supersedes #348


Itxaka:
So I guess this brings some advantages? I don't get it sorry, because it looks the same. Not saying that it's wrong, I don't know the difference between what's in there and this change :)

mauromorales:
@Itxaka the previous API got deprecated, and I think this is the new API but the test that used to be on the pkg is no more so I'm not sure if it translates 1 to 1

Itxaka (approved):

PR #358, opened 2024-05-29
Apply go mod tidy -compat=1.19 during release

mauromorales:
fixes kairos-io/kairos#2589

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -29,3 +29,8 @@ changelog:
       - '^docs:'
       - '^test:'
       - '^Merge pull request'
+env:
+  - GOSUMDB=off 
+before:
+  hooks:
+    - go mod tidy -compat=1.19
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -30,7 +30,7 @@ changelog:
       - '^test:'
       - '^Merge pull request'
 env:
-  - GOSUMDB=off 
+  - GOSUMDB=sum.golang.org
 before:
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -29,3 +29,8 @@ changelog:
       - '^docs:'
       - '^test:'
       - '^Merge pull request'
+env:
+  - GOSUMDB=off
```
jimmykarily, line 33: why?

mauromorales: good question, I don't know but otherwise it fails with GOSUMDB not found, I think it's part of the change between go 1.19 and 1.22, and since our mod file are 1.19 based then it misses some of the configuration and you have to tune it off

Itxaka: ummm this is weird, GOSUMDB should just point to the sum address like so `GOSUMDB='sum.golang.org'`

Itxaka: can we maybe set

```
gomod:
  env:
    - GOSUMDB=sum.golang.org
 ```

Comment (mauromorales) [reply .goreleaser.yaml:33]:
```suggestion
  - GOSUMDB=sum.golang.org
```

jimmykarily (approved):

.goreleaser.yaml
Itxaka: how this made it work? lol I would expect that to be the default. Maybe goreleaser using a different empty go env?

PR #362, opened 2024-06-04
Ensure only numeric part of systemd version is converted to int

ZauberNerd:
some distributions might include pre-release or release-candidate versions, e.g. debian:testing currently ships
"systemd 256~rc3 (256~rc3-7)".
This breaks `strconv.Atoi(systemdVersion)` when it receives the tilde character.

This change introduces a regular expression to ensure we only use the numeric part of the systemd version.


Itxaka (approved):
good catch!

mauromorales:
@ZauberNerd good catch, thanks for fixing 🙌

PR #372, opened 2024-06-06
Copy any found sysextensions into active+passive efi dir

Itxaka:


jimmykarily:
What I see looks good but can this be merged as it is? I mean, I guess more things are missing to implement the signed extensions feature but this at least won't break anything, right?

jimmykarily (approved):

jimmykarily:
Oh, also it will need testing but I guess we can go with and e2e test in kairos.

Itxaka:
> Oh, also it will need testing but I guess we can go with and e2e test in kairos.

I think it should be ok to test it here as part of the agent tests :D

PR #382, opened 2024-06-18
Skip dirs in copyArtifactSetRole and replace only basename

jimmykarily:
otherwise we get the error (on `kairos-agent upgrade` in uki mode):

```
panic: open /efi/EFI/kairos/passive.efi.extra.d: is a directory
```

and we may replace parts of the path that we shouldn't.

### DIFF
--- a/pkg/uki/common_test.go
+++ b/pkg/uki/common_test.go
@@ -1,0 +1,64 @@
+package uki
+
+import (
+	"bytes"
+	"os"
+
+	cnst "github.com/kairos-io/kairos-agent/v2/pkg/constants"
+	fsutils "github.com/kairos-io/kairos-agent/v2/pkg/utils/fs"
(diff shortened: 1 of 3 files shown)


pkg/uki/common_test.go
```diff
@@ -0,0 +1,69 @@
...
+			err = copyArtifactSetRole(fs, "/", "active", "passive", logger)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("replaces only the base file name", func() {
+			err = copyArtifactSetRole(fs, "/other", "other", "newother", logger)
+			Expect(err).ToNot(HaveOccurred())
+
+			glob, _ := fs.Glob("/other/*")
+			fmt.Printf("glob = %+v\n", glob)
```
jimmykarily, line 61: debug output, will remove

Itxaka (approved):

PR #385, opened 2024-06-19
Bump kcrypt and adapt hooks

Itxaka:
Now we can pass the logger directly so we see the logs as usual Also set systemd debug mode before running the commands so we can see the debug output in case of failrue


go.sum
```diff
@@ -321,6 +321,10 @@ github.com/kairos-io/kairos-sdk v0.2.3 h1:QXqjVtOm5Ki4N1diI+H6ydrStFhDHDbDkK25HJ
 github.com/kairos-io/kairos-sdk v0.2.3/go.mod h1:wbQNMckCGT/kKd+tqhAZo4thDBPWxpcKaymWJmRqSUg=
 github.com/kairos-io/kcrypt v0.11.2 h1:qu00WGDKvOn6tm9xaWho3RRpo0iRQBzZa9M/hnXhLbs=
 github.com/kairos-io/kcrypt v0.11.2/go.mod h1:znIdTYzsTtePMDA2v5bDJeikqxIb0W6YJMH77TU2Phs=
```
jimmykarily, line 323: these need `go mod tidy` to go away?

Itxaka: I have given up trying to understand go mod whatever. there is no logic to it

jimmykarily (approved):

Itxaka:
To test the if luks1 can be unlocked with newer kcrypt version:

 - installed with https://github.com/kairos-io/kairos/releases/download/v3.1.0-alpha1/kairos-ubuntu-24.04-core-amd64-generic-v3.1.0-alpha1.iso which has the previous kcrypt versions which defaults to luks1
 - installed with offline encryption
 - rebooted to active and confirmed that disk is encrypted and unlocked correctly
 - confirmed the luks header
image
 - Confirmed that the immucore version (https://github.com/kairos-io/immucore/tree/v0.2.0) that unlocks the device has the new kcrypt(https://github.com/kairos-io/kcrypt/tree/v0.11.2) with the new methods

So basically the luks version does not affect the unlocking.


To test the new luks2 version by default:

 - Built a new image with kairos-agent pointing to this branch
```
earthly +iso \
            --FLAVOR=ubuntu \
            --FLAVOR_RELEASE=24.04 \
            --FAMILY=ubuntu \
            --MODEL=generic \
            --VARIANT=core \
            --BASE_IMAGE=ubuntu:24.04 --KAIROS_AGENT_DEV_BRANCH=bump_kcrypt2
```
 - Installed the iso with offline encryption
 - rebooted to active and confirmed that disk is encrypted and unlocked correctly
 - confirmed the luks header
image
 - confirmed that immucore+agent had the proper versions of kcrypt



So basically it should not affect anything to change the default to luks2

PR #422, opened 2024-07-23
Introduce allow-discards config option

Itxaka:
This makes the luks encrypted partitions be created with the --allow-discards flag, which allows the underlying systems to be mounted with the discard option, which supposedly its better for nvme drives


Itxaka:
not needed, we gonna do it directly and then on mounting decide the discard option

PR #426, opened 2024-07-25
Bump golang deps to get newest entities and fix tredoe/osutil deleted tag

jimmykarily:
This should help fix some issues reported by `govulncheck ./...`


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kairos-io/kairos-agent/v2
 
-go 1.19
+go 1.22.5
```
mauromorales, line 3: can we keep this as 1.22? to always get the latest patch?

jimmykarily: I based on an older `main` and it turns out we already have `1.22.5` in `main`: https://github.com/kairos-io/kairos-agent/blob/f2dc6bdbb5a3cc5021ce032dc3889e6e1db0667f/go.mod#L3

I will rebase, reapply all and use 1.22

jimmykarily:
Stupid me. The needed bumps are already fixed in main. Closing this.

PR #429, opened 2024-07-26
2710 handle uki list releases

jimmykarily:
Consume: https://github.com/kairos-io/kairos-sdk/pull/262/files

Result looks like this: 
image


mauromorales (approved):
lgtm

PR #439, opened 2024-08-06
Introduce partitioner v2

Itxaka:
All in pure golang, expect the filesystem formatting

Substitutes the old one that shelled out to parted.

Manually tested and seems to work as expected


Itxaka:
Also works on BIOS installs :D

pkg/partitionerv2/mkfs.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package partitioner
+package partitionerv2
```
mauromorales, line 17: why can't we leave this one as partitioner ?

Itxaka: because Im stupid? 

I had them along together in case we wanted to reverse it :D

pkg/action/bootentries_test.go
```diff
@@ -664,12 +662,10 @@ var _ = Describe("Bootentries tests", Label("bootentry"), func() {
 				Expect(err).ToNot(HaveOccurred())
 				err = fs.WriteFile("/etc/kairos/branding/grubmenu.cfg", []byte("whatever whatever --id kairos3 {"), os.ModePerm)
 				Expect(err).ToNot(HaveOccurred())
+				Expect(fs.Mkdir("/oem", os.ModePerm)).To(Succeed())
 
 				err = SelectBootEntry(config, "kairos")
 				Expect(err).ToNot(HaveOccurred())
-				Expect(runner.IncludesCmds([][]string{
-					{"grub2-editenv", "/oem/grubenv", "set", "next_entry=kairos"},
-				})).ToNot(HaveOccurred())
```
jimmykarily, line 672: The test wasn't checking much but now it checks even less. Is it possible to check the actual result of setting the default entry using the config.Fs somehow? Some file must have changed, right?

pkg/elemental/elemental_test.go
```diff
@@ -426,16 +389,18 @@ var _ = Describe("Elemental", Label("elemental"), func() {
 				install.PartTable = v1.GPT
 				install.Firmware = v1.EFI
 				install.Partitions.SetFirmwarePartitions(v1.EFI, v1.GPT)
-				Expect(el.PartitionAndFormatDevice(install)).To(BeNil())
-				Expect(runner.MatchMilestones(append(efiPartCmds, partCmds...))).To(BeNil())
+				Skip("Not ready")
```
jimmykarily, line 392: what's missing?

jimmykarily (approved):

pkg/elemental/elemental.go
```diff
@@ -51,71 +55,67 @@ func (e *Elemental) FormatPartition(part *v1.Partition, opts ...string) error {
...
-func (e *Elemental) PartitionAndFormatDevice(i v1.SharedInstallSpec) error {
-	disk := partitioner.NewDisk(
-		i.GetTarget(),
-		partitioner.WithRunner(e.config.Runner),
-		partitioner.WithFS(e.config.Fs),
-		partitioner.WithLogger(e.config.Logger),
-	)
-
-	if !disk.Exists() {
+func (e *Elemental) PartitionAndFormatDevicev2(i v1.SharedInstallSpec) error {
```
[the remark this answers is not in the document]
mauromorales: yup also would go with the (without v2 name) or rename if there's a reason to distinguish them

pkg/elemental/elemental_test.go
```diff
@@ -334,19 +350,16 @@ var _ = Describe("Elemental", Label("elemental"), func() {
...
-		var failPart bool
 		var install *v1.InstallSpec
 		var err error
 
 		BeforeEach(func() {
 			cInit = &v1mock.FakeCloudInitRunner{ExecStages: []string{}, Error: false}
 			config.CloudInitRunner = cInit
 			config.Install.Device = "/some/device"
-			el = elemental.NewElemental(config)
+			//el = elemental.NewElemental(config)
```
mauromorales, line 362: can be removed i guess

pkg/partitionerv2/mkfs.go
mauromorales: gotcha, I think if we rollback I'd rather go with rolling back the merged commit, and the only case I'd keep them side by side is if there was a special case and we needed to still use the previous command, or if you were migrating them on batches or something like that

mauromorales (approved):
I made just very tiny styling observations, nothing that should block this from merging

pkg/action/bootentries_test.go
Itxaka: done

pkg/elemental/elemental.go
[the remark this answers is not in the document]
Itxaka: done

pkg/elemental/elemental_test.go
Itxaka: done

pkg/elemental/elemental_test.go
Itxaka: mocking the partition stuff somehow or point it to a file instead

pkg/partitionerv2/mkfs.go
Itxaka: done

Itxaka:
only thing missing is mocking or checking the elemental tests for PartitionAndFormatDevice

pkg/elemental/elemental.go
```diff
@@ -58,64 +62,63 @@ func (e *Elemental) FormatPartition(part *v1.Partition, opts ...string) error {
...
+	}
+	err = disk.Close()
+	if err != nil {
+		e.config.Logger.Errorf("Close disk: %s", err)
+	}
+	// Sync changes
+	syscall.Sync()
+	// Trigger udevadm to refresh devices
+	_, _ = e.config.Runner.Run("udevadm", "trigger")
+	_, _ = e.config.Runner.Run("udevadm", "settle")
```
mauromorales, line 104: is it worth putting the output of these calls? or at least something in the log to say doing trigger now, doing settle now, etc?

pkg/utils/grub.go
```diff
@@ -227,17 +230,30 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
...
+
+	for _, k := range keys {
+		if len(vars[k]) > 0 {
+			b.WriteString(fmt.Sprintf("%s=%s\n", k, vars[k]))
 		}
 	}
-	return nil
+
+	// https://www.gnu.org/software/grub/manual/grub/html_node/Environment-block.html
+	// It has to be exactly 1024bytes, if its not it has to be filled with #
```
mauromorales, line 251: hehe 😅

pkg/utils/utils_test.go
```diff
@@ -881,27 +883,21 @@ var _ = Describe("Utils", Label("utils"), func() {
 		})
 		Describe("SetPersistentVariables", func() {
 			It("Sets the grub environment file", func() {
-				grub := utils.NewGrub(config)
-				Expect(grub.SetPersistentVariables(
-					"somefile", map[string]string{"key1": "value1", "key2": "value2"},
+				Expect(utils.SetPersistentVariables(
+					"/tmp/caca", map[string]string{"key1": "value1", "key2": "value2"},
```
mauromorales, line 887: 🤣  maybe a less funny name?

mauromorales (approved):
some minor observations from my side

pkg/elemental/elemental.go
Itxaka: I dont think so, the calls actually return nothing, no output from the commands so there is no value on it I think. I never ever seen those error out as well xD

pkg/utils/utils_test.go
Itxaka: wopsie doopsie

pkg/elemental/elemental.go
jimmykarily: Since this function already returns an error, I don't see why we shouldn't handle the error here. It may never happen but if it does, it makes it a real pita to debug with zero-output and error being skipped. To my experience, skipping errors never proved to be a good idea.

pkg/action/install_test.go
```diff
@@ -94,41 +95,30 @@ var _ = Describe("Install action tests", func() {
 	Describe("Install Action", Label("install"), func() {
 		var device, cmdFail string
 		var err error
-		var cmdline func() ([]byte, error)
 		var spec *v1.InstallSpec
 		var installer *action.InstallAction
 
 		BeforeEach(func() {
-			device = "/some/device"
+			os.RemoveAll("/tmp/test.img")
```
jimmykarily, line 102: oh, this call is now redundant (we call it again in the next line)

pkg/action/install_test.go
```diff
@@ -94,41 +95,30 @@ var _ = Describe("Install action tests", func() {
...
+			Expect(os.RemoveAll("/tmp/test.img")).ToNot(HaveOccurred())
+			// at least 2Gb in size as state is set to 1G
+			_, err = diskfs.Create("/tmp/test.img", 2*1024*1024*1024, diskfs.Raw, 512)
+			Expect(err).ToNot(HaveOccurred())
+			device = "/tmp/test.img"
```
jimmykarily, line 107: Move this definition above the `RemoveAll` and use it in `RemoveAll` too?

pkg/action/install_test.go
```diff
@@ -214,6 +196,10 @@ var _ = Describe("Install action tests", func() {
 			installer = action.NewInstallAction(config, spec)
 		})
 		AfterEach(func() {
+			if CurrentSpecReport().Failed() {
+				GinkgoWriter.Printf(memLog.String())
+			}
+			Expect(os.RemoveAll("/tmp/test.img")).ToNot(HaveOccurred())
```
jimmykarily, line 202: Maybe worth making it a var and setting the value at the top level `BeforeEach`. It will make it easier to change the location of that file if we need to in the future (e.g. if tests start conflicting with each other because of the non-random name).

pkg/constants/constants.go
```diff
@@ -203,3 +204,6 @@ func BootTitleForRole(role, title string) (string, error) {
 		return "", errors.New("invalid role")
 	}
 }
+
+// DiskUUID is the static UUID for main disk identification
+var DiskUUID = uuid.NewV5(uuid.NamespaceURL, "KAIROS_DISK").String()
```
jimmykarily, line 209: Consumers of this will call it with something like `constants.DiskUUID` which is a bit misleading. This is not a real constant, it changes value on every invocation of the `kairos-agent` right? It could lead to bugs if someone uses this to find the main disk because it won't return the same value every time you call it.

Maybe make it a method somewhere with a clear name? (e.g. `GenerateDiskUUID`) ?

pkg/action/install_test.go
Itxaka: ah yes, this was because I had some files from manual tests around and diskfs would fail if the file existed. Will remove!

pkg/action/install_test.go
Itxaka: yep, good point

pkg/constants/constants.go
Itxaka: nope, its always the same, it generates a predictable uuid based on the type and text. I know that usually uiuid is for random stuff lol, but this makes it so the disk uuid will always be the same, so in the future we can even identify the kairos install disk by this uuid.

Same with the partitions, they are now all set to a predictable uuid based on their label name so again, we could identify the kairos partitions via uuid directly :D

jimmykarily: gotcha. I wonder what these magic hashes are supposed to be:

```
	NamespaceURL  = Must(FromString("6ba7b811-9dad-11d1-80b4-00c04fd430c8"))
```

Itxaka: https://datatracker.ietf.org/doc/html/rfc4122#appendix-C

The answer is always on the spec :P

For some reason all of them have the same UUID so its kind of moot which one to use lol

Itxaka:
Lets gooooo

PR #467, opened 2024-08-24
Fix autocalculating size image for upgrade action

dnugmanov:
Fix: https://github.com/kairos-io/kairos/issues/2818


dnugmanov:
FYI: I noticed that some tests were triggered. I will address the issues and sign off on the commit once the fixes are implemented.

pkg/config/spec.go
```diff
@@ -276,28 +276,49 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
 			recMnt = constants.TransitionDir
 		}
 
+		upgradeRecoverySystemUri, err := cfg.Query("upgrade.\"recovery-system\".uri")
```
jimmykarily, line 279: I'm trying to fix the tests currently. I think this should be `install.recovery-system.uri` but I'm not sure yet. On it.

jimmykarily:
I started fixing the tests but then did some refactoring because it turns out we don't need to to `cfg.Query` at all. We have access through the object itself.

I'm closing this in favor or: https://github.com/kairos-io/kairos-agent/pull/499

dnugmanov:
> I started fixing the tests but then did some refactoring because it turns out we don't need to to `cfg.Query` at all. We have access through the object itself.
> 
> I'm closing this in favor or: #499

Hey @jimmykarily,

Thanks for jumping in to help with this bug! I was hesitant to use `cfg.Query` since it’s a call with gojq, but I couldn’t locate `cfg.Install.Recovery.Source`. I’m glad we found the best solution in the end.

PR #499, opened 2024-08-28
Fix image size calculation

jimmykarily:
Supersedes https://github.com/kairos-io/kairos-agent/pull/467/files


pkg/config/spec.go
```diff
@@ -277,13 +277,18 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
 			recMnt = constants.TransitionDir
 		}
 
+		recoverySrc := cfg.Install.Recovery.Source
```
jimmykarily, line 280: I wonder if `cfg.Install` or `cfg.Install.Recovery` could be `nil` at this point. That would panic.

jimmykarily: Added  a check

jimmykarily:
With kairos-agent from this branch, I still see this text:

```
2024-08-28T12:43:40Z INF Creating file system image /run/initramfs/cos-state/cOS/transition.img with size 3072Mb
```

on the ugprade logs. Either the text is wrong or it didn't work as expected.

mauromorales (approved):

internal/agent/upgrade.go
```diff
@@ -238,12 +238,14 @@ func upgradeUki(source string, dirs []string, upgradeEntry string, strictValidat
 // ExtraConfigUpgrade is the struct that holds the upgrade options that come from flags and events
 type ExtraConfigUpgrade struct {
```
jimmykarily, line 239: We take the cli args and we make a new type of config (this one). Then we serialize this thing to a json (or YAML?) string and we merge it with the rest of the config. The keys of the resulting string didn't match those of the Config we are trying to build thus the setting only made it to the raw Config (which we store as a separate key). That's why the original PR (see the first commit in this one) had to use `cfg.Query` to get to this key.

With the changes here, the actual fields of the Config get populated correctly.

pkg/config/spec.go
```diff
@@ -324,11 +334,18 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
...
+	// which makes config.upgrade.entry be "recovery"
+	// Another way is by setting the cli arg "boot-entry" which set it to the
+	// specified value.
+	// Lastly, user can set "upgrade.recovery: true" in the kairos config, which
+	// should result in entry being "recovery".
 	entry := ""
 	_, ok := cfg.Config["upgrade"]
 	if ok {
+		// check value from --recovery and --boot-entry
+		entry, _ = cfg.Config["upgrade"].(collector.Config)["entry"].(string)
```
jimmykarily, line 347: We were skipping this but not tests were broken. This means we need to add a test to make sure we don't break it again.

jimmykarily: It seems that this wasn't needed because of the unmarshallFullSpec below: https://github.com/kairos-io/kairos-agent/pull/499/files#diff-53b724197bc14c3828ce6725322b892cb788860b0526184b63d7d114edb60941R370

`generateUpgradeConfForCLIArgs` has added the field `upgrade.entry` (see `ExtraConfigUpgrade`) which is marshalled into a yaml and passed to the config collector (as a string).

Let's check if this is still working as intended.

internal/agent/upgrade.go
```diff
@@ -146,8 +146,8 @@ func generateUpgradeConfForCLIArgs(source, upgradeEntry string) (string, error)
 	// have access to that yet, we just set both uri values which shouldn't matter
 	// anyway, the right one will be used later in the process.
 	if source != "" {
-		upgradeConfig.Upgrade.RecoverySystem.URI = source
-		upgradeConfig.Upgrade.System.URI = source
+		upgradeConfig.Install.RecoverySystem.URI = source
```
Itxaka, line 149: this feels wrong. This has worked with cli flags since forever, so it doesnt seem like it would be the source of the issue. The end config/spec has the proper sources so it seems wrong to me that this would be broken here?

Are we sure this doesnt break anything? IIRC our test use the cli flags to upgrade by setting a source, so this seems to be working as expected or our tests are fully broken.

I dont have anything else to proof this other than a feeling that if source cli flags work, this migth be working but I could be wrong...

jimmykarily:
We discussed this in the planning meeting and we agreed this part of the code needs some refactoring but for now we'll go with this fix instead: https://github.com/kairos-io/kairos-agent/pull/508/files

I will try to move the new test from this PR to the #508 one to have it as an additional check.

jimmykarily:
I created this PR for the test and the additional fix: https://github.com/kairos-io/kairos-agent/pull/509
Closing this one.

PR #508, opened 2024-09-02
Fix image size calculation on upgrade

Itxaka:
Because upgrade its a bit different than install, where we know where to check for the source easily, we cannot do the size calculation and then unmarshall the spec as the sources will be empty at that point.

Instead, unmarshall the full config into the spec and then calculate the size of the given source. So this should work for config file and flags the same and for normal upgrade and recovery upgrade, it just needs to fully load and parse the sources first in order to be able to calculate the size

Fixes https://github.com/kairos-io/kairos/issues/2818

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -346,13 +346,15 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
 		State:      installState,
 	}
 
-	setUpgradeSourceSize(cfg, spec)
-
+	// Unmarshall the config into the spec first so the active/recovery gets filled properly from all sources
 	err = unmarshallFullSpec(cfg, "upgrade", spec)
 	if err != nil {
 		return nil, fmt.Errorf("failed unmarshalling the full spec: %w", err)
 	}
-
+	err = setUpgradeSourceSize(cfg, spec)
+	if err != nil {
+		return nil, fmt.Errorf("failed calculating size: %w", err)
+	}
 	return spec, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/config/spec.go
```diff
@@ -367,7 +369,7 @@ func setUpgradeSourceSize(cfg *Config, spec *v1.UpgradeSpec) error {
 		targetSpec = &(spec.Active)
 	}
 
-	if targetSpec.Source.IsEmpty() {
+	if targetSpec.Source != nil && targetSpec.Source.IsEmpty() {
```
Itxaka, line 372: this is an extra check nice to add if you hack around this and suddenly pass an empty source so it doesnt fail as hard :D

Itxaka:
good failure, ti confirms that now its trying to calculate the size instead of just failing lol:

```
                 +test *failed* |   [FAILED] in [It] - /build/pkg/config/spec_test.go:589 @ 09/02/24 11:25:21.404
                 +test *failed* | • [FAILED] [0.284 seconds]
                 +test *failed* | Types Config Config from cloudconfig [It] Reads properly the cloud config for upgrade [types, config, cloud-config]
                 +test *failed* | /build/pkg/config/spec_test.go:580
  
                 +test *failed* |   [FAILED] Unexpected error:
                 +test *failed* |       <*errors.errorString | 0xc0000a9210>: 
                 +test *failed* |       failed initializing spec:
```

Sending a fixed test to check it.

mauromorales (approved):

jimmykarily:
Let's merge this and I will rebase [my PR](https://github.com/kairos-io/kairos-agent/pull/499/files) on top because it includes another fix for the `--recovery` flag and an additional test which we decided to keep.

jimmykarily (approved):

PR #525, opened 2024-09-10
Display net info on QR code page

Itxaka:
image

Fixes https://github.com/kairos-io/kairos/issues/2826


internal/agent/install.go
```diff
@@ -164,6 +167,7 @@ func Install(sourceImgURL string, dir ...string) error {
 
 	if tk != "" {
 		qr.Print(tk)
+		displayInfo(agentConfig)
```
jimmykarily, line 170: I can't follow the code. When does this happen?

Itxaka: on images with provider during boot it triggers the install and shows the QR code and waits for an event. Its weird, yeah

PR #527, opened 2024-09-10
Fail if remote url address doesnt exist

Itxaka:
We are not failing during manual install and with a remote url if that file doesnt exists, which can lead to users thinking that they are installing from a remote url but they are not, they are just using defaults.

This will make it so it fails if the remote url doesnt exists.

Fixes https://github.com/kairos-io/kairos/issues/2609


jimmykarily (approved):
```
root@localhost:~# ./kairos-agent manual-install https://www.google.com/test
configuration file not found in remote address
root@localhost:~# ./kairos-agent manual-install https://asdf
Head "https://asdf": dial tcp: lookup asdf on 10.0.2.3:53: no such host
```
LGTM

PR #530, opened 2024-09-11
Improve get partitions and reset spec

Itxaka:
- do not fill mountpoints if we dont need them
 - try to unmount partitions before we format them
 - only warn about missing partitions if we are asking to format them
 - getpartitions now tries to find the mountpoint by label as well

The main issue here was that a partition mounted as /dev/disk/by-label/LABEL would not be found properly by ghw library, so the mountpoint will always be empty

This patch introduces an extra check on the GetAllPartitions functions that if the mountpoint its empty it will try to infer if it was mounted by label instead and gather the mountpoint

Only reset should be affected by this, as we only format in that case and we only need the mountpoint during the reset to be able to unmount it beforehand. In upgrade and install we dont really need that so its not implemented there.

Fixes: https://github.com/kairos-io/kairos/issues/2836


pkg/config/spec.go
```diff
@@ -438,34 +438,18 @@ func NewResetSpec(cfg *Config) (*v1.ResetSpec, error) {
...
 		// We could have oem in lvm which won't appear in ghw list
 		ep.OEM = partitions.GetPartitionViaDM(cfg.Fs, constants.OEMLabel)
 	}
 
-	if ep.OEM == nil {
-		cfg.Logger.Warnf("no OEM partition found")
-	}
-
 	// Persistent partition is not a hard requirement
-	if ep.Persistent != nil {
```
Itxaka, line 457: removed clutter here and above, not really needed.

pkg/config/spec.go
```diff
@@ -522,6 +506,25 @@ func NewResetSpec(cfg *Config) (*v1.ResetSpec, error) {
 		return nil, fmt.Errorf("failed unmarshalling the full spec: %w", err)
 	}
 
+	if ep.OEM == nil && spec.FormatOEM {
+		cfg.Logger.Warnf("no OEM partition found, wont format it")
```
Itxaka, line 510: only warn if we are actually formatting the partition

pkg/config/spec.go
```diff
@@ -522,6 +506,25 @@ func NewResetSpec(cfg *Config) (*v1.ResetSpec, error) {
...
+	}
+
+	if ep.Persistent == nil && spec.FormatPersistent {
+		cfg.Logger.Warnf("no Persistent partition found, wont formate it")
+	}
+
+	// If we mount partitions by the /dev/disk/by-label stanza, their mountpoints wont show up due to ghw not
```
Itxaka, line 517: this is the real change that fixes this + the unmount

pkg/elemental/elemental.go
```diff
@@ -47,14 +47,7 @@ func NewElemental(config *agentConfig.Config) *Elemental {
 
 // FormatPartition will format an already existing partition
 func (e *Elemental) FormatPartition(part *v1.Partition, opts ...string) error {
-	var name string
-	// Nice display name for logs
```
Itxaka, line 51: not really neede, we can use the label directly?

pkg/utils/partitions/getpartitions.go
```diff
@@ -61,6 +64,59 @@ func GetAllPartitions() (v1.PartitionList, error) {
 	return parts, nil
 }
 
+// GetMountPointByLabel will try to get the mountpoint by using the label only
+// so we can identify mounts the have been mounted with /dev/disk/by-label stanzas
+func GetMountPointByLabel(label string) string {
```
Itxaka, line 69: at this point, reimplementing what GHW does its just a few lines more and we can adapt to our own specifics...

pkg/config/spec.go
```diff
@@ -522,6 +506,25 @@ func NewResetSpec(cfg *Config) (*v1.ResetSpec, error) {
 		return nil, fmt.Errorf("failed unmarshalling the full spec: %w", err)
 	}
 
+	if ep.OEM == nil && spec.FormatOEM {
+		cfg.Logger.Warnf("no OEM partition found, wont format it")
+	}
+
+	if ep.Persistent == nil && spec.FormatPersistent {
+		cfg.Logger.Warnf("no Persistent partition found, wont formate it")
```
jimmykarily, line 514:
```suggestion
		cfg.Logger.Warnf("no Persistent partition found, won't format it")
```

pkg/config/spec.go
jimmykarily, line 510:
```suggestion
		cfg.Logger.Warnf("no OEM partition found, won't format it")
```

pkg/config/spec.go
jimmykarily: It's not clear why it's not needed but I trust you :). It seems to be setting the default for some reason.

jimmykarily (changes requested):
Since this is the agent, this may not be covered by any tests here but we should probably add a test in the e2e suite to make sure we don't break it again.

Itxaka:
> Since this is the agent, this may not be covered by any tests here but we should probably add a test in the e2e suite to make sure we don't break it again.

We have a reset test already in the e2e suite of kairos. We dont have a test for reset-oem as that would need extra work to support it, but we should do one. Something like reset-oem + after-reset-chroot stage to create a new cloud-config so on restart it can login :D

Also there is unit tests for this I think which should cover this changes.

jimmykarily:
> > Since this is the agent, this may not be covered by any tests here but we should probably add a test in the e2e suite to make sure we don't break it again.
> 
> We have a reset test already in the e2e suite of kairos. We dont have a test for reset-oem as that would need extra work to support it, but we should do one. Something like reset-oem + after-reset-chroot stage to create a new cloud-config so on restart it can login :D
> 
> Also there is unit tests for this I think which should cover this changes.

If it was covered, shouldn't those tests be broken?

pkg/action/reset.go
```diff
@@ -114,6 +114,7 @@ func (r ResetAction) Run() (err error) {
 	defer func() { err = cleanup.Cleanup(err) }()
 
 	// Unmount partitions if any is already mounted before formatting
+	// TODO: Is this needed???
```
jimmykarily, line 117: Who can answer that? :D

jimmykarily (approved):

Itxaka:
> > > Since this is the agent, this may not be covered by any tests here but we should probably add a test in the e2e suite to make sure we don't break it again.
> > 
> > We have a reset test already in the e2e suite of kairos. We dont have a test for reset-oem as that would need extra work to support it, but we should do one. Something like reset-oem + after-reset-chroot stage to create a new cloud-config so on restart it can login :D
> > 
> > Also there is unit tests for this I think which should cover this changes.
> 
> If it was covered, shouldn't those tests be broken?

Only the normal reset ones, we do not test the reset OEM part :)

That will change hopefully and we will see this patch fixing the test I hope?

pkg/action/reset.go
Itxaka: tests probably!

PR #535, opened 2024-09-12
Move to use our ghw clone

Itxaka:


Itxaka:
sdk updated, not blocked anymore

pkg/types/v1/config.go
```diff
@@ -220,25 +220,11 @@ func (r *EmptySpec) Sanitize() error {
...
-	Path            string   `yaml:"-"`
-	Disk            string   `yaml:"-"`
-}
-
-type PartitionList []*Partition
-
 // GetByName gets a partitions by its name from the PartitionList
-func (pl PartitionList) GetByName(name string) *Partition {
-	var part *Partition
+func GetByName(name string, partitionList types.PartitionList) *types.Partition {
```
jimmykarily, line 224: The name was kept generic because the receiver was a PartitionList. Now that it doesn't have a receiver a call to this will simply look like `GetByName` which is very generic and doesn't reveal anything about what is there to "Get".

Maybe `GetPartitionByName` ?

Alternatively, can we keep the receiver (using the new type)?

jimmykarily: same this below in `GetByLabel`

jimmykarily (approved):

pkg/types/v1/config.go
Itxaka: ah yes, good catch, I forgot about this tbh, it should probably also be moved to utils directly, makes no sense to have it here. Thanks for the reminder!

Itxaka: did merge both into a `getPartitionByNameOrLabel` function instead. Keeping it in the same file as its only use there

jimmykarily (approved):

jimmykarily:
Tests broke

Itxaka:
> Tests broke

test fixed :P

PR #537, opened 2024-09-13
Fix upgrade size calculation inside k8s

Itxaka:
As we use the suc-upgrade script to upgrade inside k8s containers, the host rootfs is mounted under `$HOST_DIR`. The upgrade image fs is actually the rootfs of the container, so its mounted on `/`

The upgrade calculation was going over this dir when calculating the size for the upgrade, as the host dir is mounted under `/` This means that we were calling the upgrade with `--source dir:/` and that tried to go into the `/$HOST_DIR` and count that size.

This created 2 issues. The first one is that the upgraed size would be doubled as it counted the upgrade image + the host contents. The second would be that counting an installed system is a bit problematic with dirs like /tmp or /proc in which there can be temp files, broken symlinks and symlinks to self which wont work inside the container (self links point to the running system, so a self link in the mounted host would point to a different process than in the running container)

To fix this, we try to get the `$HOST_DIR` variable when calculating the sizes. The suc script sets this ALWAYS so if we are running under k8s with an upgrade, we will get it and we can skip counting that dir.

KUBERNETES_SERVICE_HOST is alway set for pods running under k8s so we
can infer that we are under k8s if that its set.

If that its true and the HOST_DIR var is not set, we default to the safe
value that system-upgrade-controller sets the host root into


Itxaka:
This requires backport to 2.13.x branch

pkg/config/spec.go
```diff
@@ -811,11 +811,30 @@ func GetSourceSize(config *Config, source *v1.ImageSource) (int64, error) {
 		size = int64(float64(size) * 2.5)
 	case source.IsDir():
 		filesVisited = make(map[string]bool, 30000) // An Ubuntu system has around 27k files. This improves performance by not having to resize the map for every file visited
-
+		// In kubernetes we use the suc script to upgrade, which mounts the host root into $HOST_DIR
```
jimmykarily, line 814:
```suggestion
		// In kubernetes we use the suc script to upgrade (https://github.com/kairos-io/packages/blob/main/packages/system/suc-upgrade/suc-upgrade.sh)
		// , which mounts the host root into $HOST_DIR
```

pkg/config/spec_test.go
```diff
@@ -693,4 +699,30 @@ var _ = Describe("GetSourceSize", func() {
...
+		sizeBefore, err := config.GetSourceSize(conf, imageSource)
+		Expect(err).To(BeNil())
+		Expect(sizeBefore).ToNot(BeZero())
+
+		Expect(os.Mkdir(filepath.Join(tempDir, "host"), os.ModePerm)).ToNot(HaveOccurred())
+		Expect(createFileOfSizeInMB(filepath.Join(tempDir, "host", "what.txt"), 200)).ToNot(HaveOccurred())
+
+		sizeAfter, err := config.GetSourceSize(conf, imageSource)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(sizeAfter).ToNot(Equal(sizeBefore))
```
jimmykarily, line 726: maybe check for equality with the actual size to avoid false positives?

jimmykarily (approved):
Good catch!

pkg/config/spec_test.go
Itxaka: I dont know how to calculate it beforehand as it seems that it doesnt exactly matches the 2 * 200Mb files so I did not want to add a random number here :D

I can still add it if you want?

Itxaka:
Tested with k8s and it works as expected!

pkg/config/spec_test.go
jimmykarily: I think we add a hardcoded value to that for safety or something. I would just print the actual value and if that is consistent, I would check against it. Just to be sure we don't return a 0 or something and the test still passes.

Itxaka: ok, done, just had to lok at how we calculate the size instead of being lazy :p

jimmykarily (approved):

PR #550, opened 2024-09-19
Show sources in config string

jimmykarily:
and remove directory which is only meant to be used by yip configs (not user configs read by the kairos-agent).

This needs to be explained in the release notes.

Needs this: https://github.com/kairos-io/kairos-sdk/pull/501

Part of: https://github.com/kairos-io/kairos/issues/2737


pkg/constants/constants.go
```diff
@@ -171,7 +172,6 @@ func GetGrubModules() []string {
 func GetConfigScanDirs() []string {
 	return []string{
 		"/oem",
-		"/system/oem",
```
Itxaka, line 174: but we need this somewhere else no? As this is used by the runstage stuff IIRC to read the stages from the whole system

pkg/action/render_template.go
```diff
@@ -26,7 +26,7 @@ func RenderTemplate(path string, config *config.Config, runtime state.Runtime) (
 
 	result := new(bytes.Buffer)
 	err = tpl.Execute(result, map[string]interface{}{
-		"Config": config.Config,
+		"Config": config.Config.Values,
```
jimmykarily, line 29: I'm not sure about this one. The test template looks like this:

```
configTest: "{{.Config.testKey | upper}}"
stateTest: "{{.State.system.os.architecture}}"
```
but given `Config` is an object of [this type](https://github.com/kairos-io/kairos-agent/blob/a8e855b9337fe92688148a6aac2281c3acc2397d/pkg/config/config.go#L124), up to know, one could use this in a template: `{{ .Config.Bundles }}` or `{{ .Config.EjectCD }}`

By only passing down the `Config.Values` to the template, we allow them to refer only to user config values instead. I'm not sure if it was every desirable to allow them to access our own object fields.

jimmykarily: The alternative would be to tell the users that user config values would be accessible as `{{ .Config.Values.testKey }}` which is a bit weird because we need to document that they also have access to a completely undocumented object from code as well (which nobody knew so far?)

pkg/constants/constants.go
jimmykarily: Hm you are right. Should we just rely on headers missing in our overlay files or should I remove the directory here: https://github.com/kairos-io/kairos-agent/blob/a8e855b9337fe92688148a6aac2281c3acc2397d/main.go#L330 ?

I mean, the whole idea is that `kairos-agent config` doesn't print values from those files, but that's already the case if we skip because of the missing header. We had one of them with a header though which makes me think that this is error prone and maybe we should skip the dir too?

Itxaka: I would drop that and have a separate function that get the YipConfigsDirs that includes all of the above.

pkg/action/render_template.go
Itxaka: well, do we want to expose in the templates internal values used for things like install and such? Do we know of the repercussions of passing a logger object or a mounter or an fs to the templating stuff?

IMO I would leave only the values in there, restrict the rest. If someone was using that then... weird.

pkg/constants/constants.go
jimmykarily: and use that in RunStage I guess?

jimmykarily: that's another kairos-sdk release. By the way, these are all irrelevant to this PR. Should we maybe do it separately? This is about bumpind the sdk to show the sources.

pkg/action/render_template.go
jimmykarily: I agree. ok let's go with this solution here then. If someone complains, then at least we'll know what the use case is. It was not documented anyway, if anyone was (ab)using it, they did it by looking at the code.

pkg/constants/constants.go
jimmykarily: I reverted the change so that we do it in another PR

Itxaka (approved):

pkg/action/render_template.go
jimmykarily: https://github.com/kairos-io/kairos-agent/pull/562

PR #559, opened 2024-09-25
Scan /oem the last one

Itxaka:
In case there is bundled configs, plus userdata, the userdata should be read at the end to be able to override any existing configs bundled with the system, as userdata is more dynamic.


nianyush (approved):

jimmykarily (approved):

Itxaka:
Requires some kind of docs to document this, so users know they can override a bundled config in an iso with userdata

mauromorales (approved):
We should add an announcement on the docs for this change

Itxaka:
closed in favor of https://github.com/kairos-io/kairos-agent/pull/562

PR #565, opened 2024-09-27
Fix reset

Itxaka:
Seems like witht he older ghw, we were not receiving all the proper partitions and mountpoints and such, so on reset we could blindily try to unmount all partitions and call it a day.

But now we get info about all of them, so doing so will even unmount the current recovery partition, where the reset source is at, thus failing the reset copy.

This patch fixes it by instead of going crazy, making sure we have an order mount/unmount of the proper places at the proper times and we restore everything as is once we are done with it

### DIFF
--- a/pkg/action/reset.go
+++ b/pkg/action/reset.go
@@ -113,24 +113,35 @@ func (r ResetAction) Run() (err error) {
 	cleanup := utils.NewCleanStack()
 	defer func() { err = cleanup.Cleanup(err) }()
 
-	// Unmount partitions if any is already mounted before formatting
-	// TODO: Is this needed???
-	err = e.UnmountPartitions(r.spec.Partitions.PartitionsByMountPoint(true, r.spec.Partitions.Recovery))
-	if err != nil {
-		return err
-	}
-
 	// Reformat state partition
-	err = e.FormatPartition(r.spec.Partitions.State)
-	if err != nil {
-		return err
-	}
+	// We should expose this under a flag, to reformat state before starting
+	// In case state fs is broken somehow
+	/*
+		if r.spec.FormatState {
+			state := r.spec.Partitions.State
+			if state != nil {
+				err = e.UnmountPartition(state)
+				if err != nil {
+					return err
+				}
+				err = e.FormatPartition(state)
+				if err != nil {
+					return err
+				}
+				// Mount it back
+				err = e.MountPartition(state)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	*/
 
 	// Reformat persistent partition
 	if r.spec.FormatPersistent {
 		persistent := r.spec.Partitions.Persistent
 		if persistent != nil {
-			err = e.UnmountPartitions(r.spec.Partitions.PartitionsByMountPoint(true, persistent))
+			err = e.UnmountPartition(persistent)
 			if err != nil {
 				return err
 			}
@@ -146,7 +157,7 @@ func (r ResetAction) Run() (err error) {
 		oem := r.spec.Partitions.OEM
 		if oem != nil {
 			// Try to umount
-			err = e.UnmountPartitions(r.spec.Partitions.PartitionsByMountPoint(true, oem))
+			err = e.UnmountPartition(oem)
 			if err != nil {
 				return err
 			}
@@ -154,22 +165,26 @@ func (r ResetAction) Run() (err error) {
 			if err != nil {
 				return err
 			}
+			// Mount it back, as oem is mounted during recovery, keep everything as is
+			err = e.MountPartition(oem)
+			if err != nil {
+				return err
+			}
 		}
 	}
-	// Mount configured partitions
-	err = e.MountPartitions(r.spec.Partitions.PartitionsByMountPoint(false, r.spec.Partitions.Recovery))
-	if err != nil {
-		return err
-	}
-	cleanup.Push(func() error {
-		return e.UnmountPartitions(r.spec.Partitions.PartitionsByMountPoint(true, r.spec.Partitions.Recovery))
-	})
 
 	// Before reset hook happens once partitions are aready and before deploying the OS image
 	err = r.resetHook(cnst.BeforeResetHook, false)
 	if err != nil {
 		return err
 	}
+
+	// Mount COS_STATE so we can write the new images
+	err = e.MountPartition(r.spec.Partitions.State)
+	if err != nil {
+		return err
+	}
+	cleanup.Push(func() error { return e.UnmountPartition(r.spec.Partitions.State) })
 
 	// Deploy active image
 	meta, err := e.DeployImage(&r.spec.Active, true)
--- a/pkg/action/reset.go
+++ b/pkg/action/reset.go
@@ -116,6 +116,9 @@ func (r ResetAction) Run() (err error) {
 	// Reformat state partition
 	// We should expose this under a flag, to reformat state before starting
 	// In case state fs is broken somehow
+	// We used to format it blindly but if something happens between this format and the new image copy down below
+	// You end up with a broken system with no active or passive and reset may not even work at that point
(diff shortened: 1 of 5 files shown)


pkg/action/reset.go
```diff
@@ -113,24 +113,35 @@ func (r ResetAction) Run() (err error) {
...
+				}
+			}
+		}
+	*/
 
 	// Reformat persistent partition
 	if r.spec.FormatPersistent {
 		persistent := r.spec.Partitions.Persistent
 		if persistent != nil {
-			err = e.UnmountPartitions(r.spec.Partitions.PartitionsByMountPoint(true, persistent))
```
jimmykarily, line 133: didn't the `excludes` flag work? (In `PartitionsByMountPoint`). Maybe that method needs fixing then?

jimmykarily: wait, we are doing the opposite now? We use to unmount "all but persistent" and now we unmount "persistent" only?

jimmykarily:
These errors would have helped us fixing it. Can we print those errors too?

```
diff --git a/pkg/utils/common.go b/pkg/utils/common.go
index abc11b5..9262fcb 100644
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -106,14 +106,17 @@ func ConcatFiles(fs v1.FS, sources []string, target string) (err error) {
        for _, source := range sources {
                sourceFile, err := fs.Open(source)
                if err != nil {
+                       fmt.Printf("err.Error() = %+v\n", err.Error())
                        break
                }
                _, err = io.Copy(targetFile, sourceFile)
                if err != nil {
+                       fmt.Printf("err.Error() = %+v\n", err.Error())
                        break
                }
                err = sourceFile.Close()
                if err != nil {
+                       fmt.Printf("err.Error() = %+v\n", err.Error())
                        break
                }
        }
```

jimmykarily:
> These errors would have helped us fixing it. Can we print those errors too?
> 
> ```
> diff --git a/pkg/utils/common.go b/pkg/utils/common.go
> index abc11b5..9262fcb 100644
> --- a/pkg/utils/common.go
> +++ b/pkg/utils/common.go
> @@ -106,14 +106,17 @@ func ConcatFiles(fs v1.FS, sources []string, target string) (err error) {
>         for _, source := range sources {
>                 sourceFile, err := fs.Open(source)
>                 if err != nil {
> +                       fmt.Printf("err.Error() = %+v\n", err.Error())
>                         break
>                 }
>                 _, err = io.Copy(targetFile, sourceFile)
>                 if err != nil {
> +                       fmt.Printf("err.Error() = %+v\n", err.Error())
>                         break
>                 }
>                 err = sourceFile.Close()
>                 if err != nil {
> +                       fmt.Printf("err.Error() = %+v\n", err.Error())
>                         break
>                 }
>         }
> ```
> 
> using some logger preferably

actually, just return the error (wrapped or not).

pkg/action/reset.go
```diff
@@ -146,30 +157,34 @@ func (r ResetAction) Run() (err error) {
...
+			if err != nil {
+				return err
+			}
 		}
 	}
-	// Mount configured partitions
-	err = e.MountPartitions(r.spec.Partitions.PartitionsByMountPoint(false, r.spec.Partitions.Recovery))
+
+	// Before reset hook happens once partitions are aready and before deploying the OS image
+	err = r.resetHook(cnst.BeforeResetHook, false)
```
jimmykarily, line 177: any reason why we are changing the logic here? What if the before-reset hook need access to the state partition? It used to have access until now (since we mounted first)

pkg/elemental/elemental.go
```diff
@@ -350,9 +350,9 @@ func (e *Elemental) DeployImage(img *v1.Image, leaveMounted bool) (info interfac
 			}
 		}
 	} else if img.Label != "" && img.FS != cnst.SquashFs {
-		_, err = e.config.Runner.Run("tune2fs", "-L", img.Label, img.File)
+		out, err := e.config.Runner.Run("tune2fs", "-L", img.Label, img.File)
 		if err != nil {
-			e.config.Logger.Errorf("Failed to apply label %s to %s", img.Label, img.File)
+			e.config.Logger.Errorf("Failed to apply label %s to %s: %s", img.Label, img.File, out)
```
jimmykarily, line 355: I think this is `[]byte`, are you sure it doesn't need `string()` around it?

pkg/action/reset.go
Itxaka: Yes, this was wrong, it worked and we were unmounting everything but persistent for some reason. We wanted the other way, before formatting persistent we want to unmount persistent.

pkg/action/reset.go
Itxaka: I dont think we should mount everything like crazy here. If the reset hook needs persistent they should mount them in there, like we do in other places like oem and such. Recovery should only have recovery and oem mounted at teh start and then leave them mounted at the end.

pkg/action/reset.go
jimmykarily: :upside_down_face:

pkg/elemental/elemental.go
Itxaka: nope, I tested it to see the actual. It seems to cast it automatically. I can add it thought

Itxaka:
> > These errors would have helped us fixing it. Can we print those errors too?
> > ```
> > diff --git a/pkg/utils/common.go b/pkg/utils/common.go
> > index abc11b5..9262fcb 100644
> > --- a/pkg/utils/common.go
> > +++ b/pkg/utils/common.go
> > @@ -106,14 +106,17 @@ func ConcatFiles(fs v1.FS, sources []string, target string) (err error) {
> >         for _, source := range sources {
> >                 sourceFile, err := fs.Open(source)
> >                 if err != nil {
> > +                       fmt.Printf("err.Error() = %+v\n", err.Error())
> >                         break
> >                 }
> >                 _, err = io.Copy(targetFile, sourceFile)
> >                 if err != nil {
> > +                       fmt.Printf("err.Error() = %+v\n", err.Error())
> >                         break
> >                 }
> >                 err = sourceFile.Close()
> >                 if err != nil {
> > +                       fmt.Printf("err.Error() = %+v\n", err.Error())
> >                         break
> >                 }
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > using some logger preferably
> 
> actually, just return the error (wrapped or not).

tbh, I would like to do this in a different PR if possible> As it can introduce new errors and I want this to fix the actual reset issue before breaking it further lol

pkg/action/reset.go
Itxaka: If you check the hooks, you can see for example that we assume that nothing is mounted and mount things explicitely and then unmount them. I think thats the correct approach, especially on recovery where we dont want to leave things around mounted.

jimmykarily (approved):

pkg/action/reset_test.go
```diff
@@ -108,22 +107,24 @@ var _ = Describe("Reset action tests", func() {
...
+						FilesystemLabel: "COS_OEM",
 						FS:              "ext4",
+						MountPoint:      "/oem",
 					},
 					{
 						Name:            "device3",
-						FilesystemLabel: "COS_PERSISTENT",
+						FilesystemLabel: "COS_RECOVERY",
 						FS:              "ext4",
+						MountPoint:      "/run/initramfs/cos-state",
```
mauromorales, line 118: @Itxaka wait, this one says recovery, but mountpoint cos-state? is this actually how it's supposed to be? if so, can you clarify?

Itxaka: Historical reasons

*Refuses to elaborate*

Itxaka: Basically to keep the ongoing theme, when booting on recovery, the recovery partition is mounted as it was the state partition in the same path as the state partition is mounted during active/passive

No other reason than that

PR #574, opened 2024-10-09
Allow installing with no users

Itxaka:
Part of https://github.com/kairos-io/kairos/issues/2921

Covers both install and interactive-install as the RunInstall method is called from all of those entrypoints.
It also improves the interactive installer to be in sync with setting no users.


jimmykarily (approved):
awesome!

PR #575, opened 2024-10-11
Check for user+admin validity before actions

Itxaka:


internal/agent/upgrade.go
```diff
@@ -64,23 +63,35 @@ func ListNewerReleases(includePrereleases bool) ([]string, error) {
 	return tagList.FullImages()
 }
 
+// TODO: Check where force and preReleases is being used? They dont seem to be used anywhere?
```
jimmykarily, line 66: Regarding `pre`, I think we used to allow the users to upgrade "blindly" without specifying the version. In that case we did a lookup for new versions and we picked the latest. I don't see us doing that anymore in the code so this must be a relic.
Regarding `force`, I don't know what it does/did.

I don't see those documented in the docs, so maybe it's time to remove the flags from `main.go` too? (in another PR)

PR #579, opened 2024-10-16
Read actual system configs during k8s upgrade

Itxaka:
Because under k8s the actual real system is mounted into a dir, the configs that we are reading by default are not the ones in the system but the ones on the upgrade container, which is running under /

This patch introduces a check to see if we are running under k8s, and if we are it tries to get the host dir where the system is mounted. then it will prepend that host dir to the config scan so we can actually scan the system configs.

Int he case we are out of k8s, this should not affect the upgrade as it will return an empty string resulting in the ususal dirs being scanned

Fixes https://github.com/kairos-io/kairos/issues/2944 and probably others lol

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -853,9 +853,9 @@ func GetSourceSize(config *Config, source *v1.ImageSource) (int64, error) {
 
 		// This is always set for pods running under kubernetes
 		_, underKubernetes := os.LookupEnv("KUBERNETES_SERVICE_HOST")
-		config.Logger.Logger.Debug().Bool("status", underKubernetes).Msg("Running under kubernetes")
 		// Try to get the HOST_DIR in case we are not using the default one
 		hostDir := k8sutils.GetHostDirForK8s()
+		config.Logger.Logger.Debug().Bool("status", underKubernetes).Str("hostdir", hostDir).Msg("Running under kubernetes host directory")
 		err = fsutils.WalkDirFs(config.Fs, source.Value(), func(path string, d fs.DirEntry, err error) error {
 			// If its empty we are just not setting it, so probably out of the k8s upgrade path
 			if hostDir != "" && strings.HasPrefix(path, hostDir) {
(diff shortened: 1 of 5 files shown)


pkg/config/spec.go
```diff
@@ -852,14 +853,9 @@ func GetSourceSize(config *Config, source *v1.ImageSource) (int64, error) {
...
-		config.Logger.Logger.Info().Bool("status", underKubernetes).Msg("Running under kubernetes")
 		// Try to get the HOST_DIR in case we are not using the default one
-		hostDir := os.Getenv("HOST_DIR")
-		// If we are under kubernetes but the HOST_DIR var is empty, default to /host as system-upgrade-controller mounts
-		// the host in that dir by default
-		if underKubernetes && hostDir == "" {
-			hostDir = "/host"
-		}
+		hostDir := k8sutils.GetHostDirForK8s()
+		config.Logger.Logger.Debug().Bool("status", underKubernetes).Str("hostdir", hostDir).Msg("Running under kubernetes host directory")
```
jimmykarily, line 858: There is no `if` anymore. This will always read "Running under kubernetes host directory" no? Or is `Bool()` a conditions on whether to print this or not?

pkg/utils/k8s/common.go
```diff
@@ -0,0 +1,27 @@
...
+	// If we are under kubernetes but the HOST_DIR var is empty, default to /host as system-upgrade-controller mounts
+	// the host in that dir by default
+	if underKubernetes {
+		if hostDirEnv != "" {
+			return hostDirEnv
+		} else {
+			return "/host"
+		}
+	} else {
+		// We return an empty string so any filepath.join does nto alter the paths
```
jimmykarily, line 24:
```suggestion
		// We return an empty string so any filepath.join does not alter the paths
```

jimmykarily (approved):

pkg/config/spec.go
Itxaka: yes, indeed, I will change the msg.

this is just to show the values of the kubernetes and hostdirectory for debug purposes as it can happen that its doing it worngly and we have no visibility onto it

pkg/utils/k8s/common.go
Itxaka: fixed

PR #582, opened 2024-10-24
Fix partitioner on disks with secotor size other than 512

Itxaka:
We were assuming all disks in the world would have a sector size of 512 but that its not true.

This patch will not force to open the disk with a hardcoded sector size and will use the sector size values of the disk to set the sizes, partition table and such.

this fixes partitions on disks with 4k sector size

Fixes https://github.com/kairos-io/kairos/issues/2964


jimmykarily (commented):
look good. I hope it works equally well too :D

richardelling (approved):

Itxaka:
can you believe I already forgot about this!

PR #587, opened 2024-11-05
Respect user defined/default sizes on upgrade

Itxaka:
We were ignoring if the user set any values in there and just overwriting it with our calculated size

Fixes https://github.com/kairos-io/kairos/issues/2985

### DIFF
--- a/pkg/action/upgrade_test.go
+++ b/pkg/action/upgrade_test.go
@@ -18,22 +18,17 @@ package action_test
 
 import (
 	"bytes"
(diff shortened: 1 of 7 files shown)


Itxaka:
```
2024-11-05T15:01:09Z DBG Calculated size (3016Mb) is less than specified/default size (8192Mb)
2024-11-05T15:01:09Z INF Setting image size to 8192Mb
```

jimmykarily (approved):

PR #591, opened 2024-11-06
Drop those stupid github and vfs deps bumps

Itxaka:
Move to latest version of those, hope we dont get more automated renovate PRs for them


go.mod
```diff
@@ -193,6 +190,7 @@ require (
 	github.com/tklauser/go-sysconf v0.3.12 // indirect
 	github.com/tklauser/numcpus v0.6.1 // indirect
 	github.com/tredoe/osutil v1.5.0 // indirect
+	github.com/twpayne/go-vfs/v4 v4.3.0 // indirect
```
jimmykarily, line 193: shouldn't this be v5? That's what you bumped to in the code.

jimmykarily (approved):

go.mod
Itxaka: indirect, so it measn something in the deps is bringing it, but IIRC the indirect deps are not bumped by renovate directly

PR #604, opened 2024-11-20
Add boot assesment for install and bootentry

Itxaka:
Part of https://github.com/kairos-io/kairos/issues/2864

The way it works:

 - On install, it runs over all the config entries and if it has no assesment it adds it by changing the filename to append +3 to it
 - On upgrade, does the same
 - On reset, does the same
 - On bootentry, it removes the assessment, if any, from the display name for presentation purposes, then before selecting the entry, reads the assesment number from the file and writes it back if needed. This is done to avoid missing retries in case there is any (so if passive was tried twice, it would keep those tries instead of restarting them)


Itxaka:
Seems to work.

During install:
 - inits the entries with a +3

During boot:
 - bless-boot triggers and changes the conf name once the system is fully up
 - other entries are untouched
 - bootentry selects the proper conf (with or without assesment) and shows the nicer entries
 - bootentry --select selects the proper conf (with or without assessment)
 - cant select ACTIVE or PASSIVE, it only allows COS or FALLBACK (check if current selector does the same)
 
```
root@kairos-o654:~# ls /efi/loader/entries/
active.conf  passive+3.conf  recovery+3.conf  statereset+3.conf
```


On upgrade and reset:
 - Updates any entries to have boot assesment if they dont have it already (active on upgrade, both on reset as they are recreated)
```
2024-11-25T17:10:36Z DBG Conf file /efi/loader/entries/active.conf has values ma
p[string]string{
  "efi": "/EFI/kairos/recovery.efi",
  "title": "Kairos recovery",
}
2024-11-25T17:10:36Z DBG Conf file /efi/loader/entries/active.conf new values ma
p[string]string{
  "efi": "/EFI/kairos/active.efi",
  "title": "Kairos recovery",
}
2024-11-25T17:10:36Z DBG Enabling boot assessment from=/efi/loader/entries/activ
e.conf to=/efi/loader/entries/active+3.conf
2024-11-25T17:10:36Z DBG Boot assessment already present in file file=/efi/loade
r/entries/passive+3.conf
2024-11-25T17:10:36Z DBG Enabling boot assessment from=/efi/loader/entries/recov
ery.conf to=/efi/loader/entries/recovery+3.conf
2024-11-25T17:10:36Z DBG Boot assessment already present in file file=/efi/loade
r/entries/statereset+3.conf
```

jimmykarily:
This output is strange:

```
2024-11-25T17:10:36Z DBG Conf file /efi/loader/entries/active.conf has values ma
p[string]string{
  "efi": "/EFI/kairos/recovery.efi",
  "title": "Kairos recovery",
}
2024-11-25T17:10:36Z DBG Conf file /efi/loader/entries/active.conf new values ma
p[string]string{
  "efi": "/EFI/kairos/active.efi",
  "title": "Kairos recovery",
}
```
how did active.conf end up with title "Kairos recovery" (you run a reset before?). When we reset from recovery, shouldn't we replace the `active.efi` with the `recovery.efi` ? If we just point active.conf to recovery.efi, if someone upgrades the recovery image, they will accidentally replace what they boot when they boot active.

I'm not sure how the above state was reached so I may be confusing things.

Itxaka:
> This output is strange:
> 
> ```
> 2024-11-25T17:10:36Z DBG Conf file /efi/loader/entries/active.conf has values ma
> p[string]string{
>   "efi": "/EFI/kairos/recovery.efi",
>   "title": "Kairos recovery",
> }
> 2024-11-25T17:10:36Z DBG Conf file /efi/loader/entries/active.conf new values ma
> p[string]string{
>   "efi": "/EFI/kairos/active.efi",
>   "title": "Kairos recovery",
> }
> ```
> 
> how did active.conf end up with title "Kairos recovery" (you run a reset before?). When we reset from recovery, shouldn't we replace the `active.efi` with the `recovery.efi` ? If we just point active.conf to recovery.efi, if someone upgrades the recovery image, they will accidentally replace what they boot when they boot active.
> 
> I'm not sure how the above state was reached so I may be confusing things.

yes, this was a reset from recovery, which I guess makes sense somehow?

pkg/utils/fs/fs.go
```diff
@@ -273,3 +273,38 @@ func Copy(fs v1.FS, src, dst string) error {
...
+	}
+
+	// Split the pattern into directory and file parts.
+	dir, file := filepath.Split(pattern)
+	if dir == "" {
+		dir = "."
+	}
+
+	// Read the directory.
+	entries, err := fs.ReadDir(dir)
```
jimmykarily, line 295: This is assuming the glob character will be in the file part. It will not work for patterns like "/mydir/*/myfile". I guess we don't care but maybe `pattern` should be called `filePattern` instead and have another argument for the directory (which will not support globing).

pkg/utils/utils_test.go
```diff
@@ -1062,4 +1063,206 @@ var _ = Describe("Utils", Label("utils"), func() {
...
+			Expect("/efi/loader/entries/test1.conf").ToNot(matchers.BeAnExistingFileFs(fs))
+			Expect("/efi/loader/entries/test3+3.conf").ToNot(matchers.BeAnExistingFileFs(fs))
+			// Should match with the +3 and the existing ones left in place
+			Expect("/efi/loader/entries/test1+3.conf").To(matchers.BeAnExistingFileFs(fs))
+			Expect("/efi/loader/entries/test2+3.conf").To(matchers.BeAnExistingFileFs(fs))
+			Expect("/efi/loader/entries/test3+1-2.conf").To(matchers.BeAnExistingFileFs(fs))
+		})
+		It("fails to write the boot assessment in non existing dir", func() {
+			err := utils.AddBootAssessment(fs, "/fake", logger)
+			Expect(err).To(HaveOccurred())
```
jimmykarily, line 1114: Better use `MatchError` to match the specific error. Just in case something else fails in the future instead of the expected error.

jimmykarily:
I wonder if it would be possible to have an e2e test for this. Unit testing doesn't really test that we indeed boot into fallback automatically. Setting this up will definitely be tricky...

jimmykarily:
I tried this branch and it seems to be doing correct things in regards to naming conf files. What we miss is:

- automatic rebooting (or maybe it already works in some cases like kernel panics?)
- sorting the boot entries in a deterministic way so that if "active" fails, we boot into "passive". Currently the next one picked was "state reset" when I run:

```bash
mount -o remount,rw /efi
/usr/lib/systemd/systemd-bless-boot bad
reboot
```
(From the [FAQ here](https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/))

We probably need to play with `sort-key` like we discussed in Slack (https://uapi-group.org/specifications/specs/boot_loader_specification/#version-order)

Merge and create tickets for the additional stuff?

pkg/utils/fs/fs.go
Itxaka: umm  good point, it wont.

Itxaka:
https://github.com/kairos-io/kairos/issues/3041
https://github.com/kairos-io/kairos/issues/3040

PR #609, opened 2024-11-27
Add sort-key during install based on the entry name

Itxaka:
image

Fixes https://github.com/kairos-io/kairos/issues/3040

### DIFF
--- a/pkg/uki/common.go
+++ b/pkg/uki/common.go
@@ -177,7 +177,7 @@ func AddSystemdConfSortKey(fs v1.FS, artifactDir string, log sdkTypes.KairosLogg
 			log.Logger.Debug().Str("path", path).Msg("Adding sort key to file")
 			conf, err := sdkutils.SystemdBootConfReader(path)
 			if err != nil {
-				log.Errorf("Error reading conf file to extract values %s: %s", path)
+				log.Errorf("Error reading conf file to extract values %s: %s", conf, path)
 			}
 			// Now check and put the proper sort key
 			var sortKey string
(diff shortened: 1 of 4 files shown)


pkg/uki/common.go
```diff
@@ -166,3 +166,47 @@ func copyFile(src, dst string) error {
...
+			log.Logger.Debug().Str("path", path).Msg("Adding sort key to file")
+			conf, err := sdkutils.SystemdBootConfReader(path)
+			if err != nil {
+				log.Errorf("Error reading conf file to extract values %s: %s", conf, path)
+			}
+			// Now check and put the proper sort key
+			var sortKey string
+			// If we have 2 different files that start with active, like with the extra-cmdline, how do we set this?
+			// Ideally if they both have the same sort key, they will be sorted by name so the single one will be first
+			// and the extra-cmdline will be second. This is the best we can do currently without making this a mess
```
jimmykarily, line 186: If the extra was meant to be the "default", it would not be set as extra cmdline. I think extra cmdlines were always meant to be alternative boot entries, not part of the "active", "passive", "recovery" dance.

jimmykarily: Having said that, if "active" fails to boot, I don't think we would want to try "active_myops" next, would we? The next thing we would want to try would be "passive". Unless I'm missing some user flow.

Itxaka: Im not sure, maybe active and active_debug should be tried one after the other?

Thats why this has a big explanation, I guess we can add this then move into a more detailed thing or explain it better?

In any case, this is just the first implementation, usually this would not matter much (the order) as we select explicit entries, so this will come later down the line when people start implementing assessment and such?

jimmykarily (approved):

PR #624, opened 2024-12-10
Fix boot selection on uki

Itxaka:
Seems like systemd-boot identifies entries by the ID which doesnt show anywhere, and that Id is the entry conf name minus the boot assessment

This means we can drop the boot assessment part from the loader.conf default entry selection

image

### DIFF
--- a/pkg/action/bootentries.go
+++ b/pkg/action/bootentries.go
@@ -133,13 +133,9 @@ func selectBootEntrySystemd(cfg *config.Config, entry string) error {
 	if err != nil {
 		return err
 	}
-	assessment, err := utils.ReadAssessmentFromEntry(cfg.Fs, bootFileName, cfg.Logger)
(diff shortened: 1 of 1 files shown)


Itxaka:
this shit is lying to me, it doesnt seem to do what it says in there, even if I set passive.conf as the entry, it will NOT select the passive....I dont get it.

pkg/action/bootentries.go
```diff
@@ -133,13 +133,9 @@ func selectBootEntrySystemd(cfg *config.Config, entry string) error {
 	if err != nil {
 		return err
 	}
-	assessment, err := utils.ReadAssessmentFromEntry(cfg.Fs, bootFileName, cfg.Logger)
```
mauromorales, line 136: I think this function is only used here and the tests so maybe it can be removed

PR #626, opened 2024-12-11
adding error message

asteurer:
Closes https://github.com/kairos-io/kairos/issues/2152


pkg/config/spec.go
```diff
@@ -397,6 +397,18 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed calculating size: %w", err)
 	}
+
+	cfg.Logger.Infof("Checking if OCI image %s exists", spec.Active.Source.Value())
+	if spec.Active.Source.IsDocker() {
```
Itxaka, line 402: I think this is good, but Im wondering if at this point the source could be empty? Im checking the code and indeed it seems that at this point, is possible that `spec.Active.Source` is an emptySrc. 

But this code is still ok, as an empty source will still return false for the Isdocker method.

What I would change, here and in the upgradeUKISpec part, is to move the info log into the if, so it only shows that log line when its an OCI source, otherwise we are printing it always, and for something other than OCI it will print that lonely line :D

pkg/config/spec.go
```diff
@@ -397,6 +397,18 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
...
+		_, err := crane.Manifest(spec.Active.Source.Value())
```
Itxaka, line 403:
```suggestion
                cfg.Logger.Infof("Checking if OCI image %s exists", spec.Active.Source.Value())
		_, err := crane.Manifest(spec.Active.Source.Value())
```

pkg/config/spec.go
```diff
@@ -397,6 +397,18 @@ func NewUpgradeSpec(cfg *Config) (*v1.UpgradeSpec, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed calculating size: %w", err)
 	}
+
+	cfg.Logger.Infof("Checking if OCI image %s exists", spec.Active.Source.Value())
```
Itxaka, line 401:
```suggestion

```

Itxaka (approved):

PR #635, opened 2024-12-17
Expand DeployImage to be more flexible

Itxaka:
Usually we create the system dirs in images by default, but that means that we cannot reuse the DeployImage for deploying random non-system images.

This fixes it by adding an extra param to create the dir structure in the created image


pkg/elemental/elemental.go
```diff
@@ -338,9 +341,11 @@ func (e *Elemental) DeployImage(img *v1.Image, leaveMounted bool) (info interfac
 		return nil, err
 	}
 	if !img.Source.IsFile() {
-		err = utils.CreateDirStructure(e.config.Fs, target)
-		if err != nil {
-			return nil, err
+		if createDirStructure {
```
jimmykarily, line 344: Let's avoid [the evil boolean arguments](https://medium.com/@MaLiN2223/my-boolean-vendetta-a2b473a64830) and create a new `DeployImageWithDir` method instead. It will run this new block in the beginning and then it will simply call `DeployImage`. What do you think?

jimmykarily: We still have deployImage accepting boolean arguments :D. With the methods you currently have, what I meant was to move this block that runs `CreateDirStructure` inside `DeployImage` itself and skip it completely from here and from `DeployImageNoDirs`. This way, only the method that want's dirs will call it and no (new) boolean arguments are needed.
Currently it runs in the middle of the `deployImage` function but it seems independent from the code above it so it could just run before the `deployImage` call. E.g.

```
func (e *Elemental) DeployImage(img *v1.Image, leaveMounted bool) (info interface{}, err error) {
  if !img.Source.IsFile() {
    err = utils.CreateDirStructure(e.config.Fs, target)
    if err != nil {
      return nil, err
    }
  }
  return e.deployImage(img, leaveMounted)
}
```

and 

```
func (e *Elemental) DeployImageNodirs(img *v1.Image, leaveMounted bool) (info interface{}, err error) {
	return e.deployImage(img, leaveMounted)
}
```

jimmykarily: we could even get rid of `leaveMounted` in a similar way but that would produce 4 different methods (for all combinations). Let's skip it for now.

jimmykarily (approved):

pkg/elemental/elemental.go
jimmykarily: discussed in Slack, it's more complex than this. Skipping for now.

PR #649, opened 2025-01-02
Use a singleton loopManager with mutex locking

Itxaka:


Itxaka:
This should fix potential issues with several calls to the loop creation in parallel failing due to race conditions

mauromorales (approved):

Itxaka:
meh, this works if there are several calls to it, i.e. from inside the agent, but if something calls the agent in parallel it wont work. This works ok for the agent but not for something like auroraboot which can run in parallel ..... Maybe adding a retry as well?

Itxaka:
> > meh, this works if there are several calls to it, i.e. from inside the agent, but if something calls the agent in parallel it wont work. This works ok for the agent but not for something like auroraboot which can run in parallel ..... Maybe adding a retry as well?
> 
> Maybe if we lock the file with something like this? https://github.com/juju/fslock If the locking fails, it means something else locked it already, so we retry with another one.

aha! good idea and thanks for making me avoid the search for something like that!

Still if we run several auroras in parallel, they are isolated in their own containers, arent they? So there is no way they can share a lock file? Unless /run is shared or something like that?

jimmykarily:
Actually that package seems to have been replaced by this one: https://github.com/juju/mutex

Looking at the code, sharing tmp inside the auroraboot pods would be enough (?) : 

https://github.com/juju/mutex/blob/11eeddb42c6c709cc089b3806526d8bd69a9327f/mutex_flock.go#L73C32-L73C39

but that's a problem to solve on the osbuilder side, right? Multiple auroraboot instances running on the same host would share tmp anyway.

Itxaka:
I tried and I dont feel I have enough neurons to process this correctly.

PR #661, opened 2025-01-14
Some small fixes here and there

Itxaka:
Including:
 - init empty vars as the type
 - fix import order
 - fix error strings
 - fix unused imports
 - remove unused functions (if needed by anything, they shoudl go into sdk)
 - fix functions comments
 - improve constants


internal/agent/reset.go
```diff
@@ -172,7 +172,9 @@ func sharedReset(reboot, unattended, resetOem bool, dir ...string) (c *config.Co
 
 	d, err := json.Marshal(r)
 	if err != nil {
-		c.Logger.Errorf("failed to marshal reset cmdline flags/event options: %s", err)
+		if &c != nil {
```
Itxaka, line 175: this was a potential nil reference

pkg/config/config.go
```diff
@@ -442,7 +424,7 @@ func scan(result *Config, opts ...collector.Option) (c *Config, err error) {
 		}
 	}
 
-	if !kc.IsValid() {
+	if kc != nil && !kc.IsValid() {
```
Itxaka, line 427: potential nil reference

pkg/config/config.go
```diff
@@ -339,20 +335,6 @@ type Bundle struct {
 	Targets    []string `yaml:"targets,omitempty"`
 }
 
-const DefaultHeader = "#cloud-config"
-
-func HasHeader(userdata, head string) (bool, string) {
```
jimmykarily, line 344: https://github.com/kairos-io/provider-kairos/blob/5b9bdda87d290893ef58626a12585f373608e985/internal/cli/token/rotate.go#L69

pkg/types/v1/image.go
```diff
@@ -29,7 +29,7 @@ type OCIImageExtractor struct{}
 
 var _ ImageExtractor = OCIImageExtractor{}
 
-func (e OCIImageExtractor) ExtractImage(imageRef, destination, platformRef string) error {
+func (e OCIImageExtractor) ExtractImage(imageRef, destination, _ string) error {
 	img, err := utils.GetImage(imageRef, utils.GetCurrentPlatform(), nil, nil)
```
jimmykarily, line 33: shouldn't we be doing some check that the platformRef matches the current platform or something like that? You already did something like that in the kairos-sdk: https://github.com/kairos-io/kairos-sdk/pull/557/files#diff-5a53c19741f79228c7414fd4f325e7d79935eb2b5c051ae2e44c0d9521542006R111

or is that not the same case?

pkg/config/config.go
Itxaka: mmmh should be implemented there on in the sdk, makes no sense to have a function here that its used bu another library lool

pkg/types/v1/image.go
Itxaka: IIRC we are using the sdk here as a backend, this methods were here becuase it was part of the agent before we put in on the sdk and then we moved but could not update or something so we leaved this stub in here. we should be using the sdk ones directly!

jimmykarily:
I fixed the failing test and rebased on main (fixed conflicts).

Itxaka:
> I fixed the failing test and rebased on main (fixed conflicts).

still failing looooool

jimmykarily:
> > I fixed the failing test and rebased on main (fixed conflicts).
> 
> still failing looooool

not anymore :p

pkg/constants/constants.go
```diff
@@ -92,45 +92,45 @@ const (
...
 	UpgradeNoSourceError           = "Could not find a proper source for the upgrade.\nThis can be configured in the cloud config files under the 'upgrade.system.uri' key or via cmdline using the '--source' flag."
 	MultipleEntriesAssessmentError = "multiple boot entries found for %s"
 	NoBootAssessmentWarning        = "No boot assessment found in current boot entry config file"
 )
 
+const (
+	_ = 1 << (10 * iota)
+	_
+	_
+	GiB
```
mauromorales, line 131: WAT? hehe I guess it just means we are not using the smaller units 😅

Itxaka: xDD

I feel weird about that. Its some kind of magic and then it turns into real GiB....

PR #679, opened 2025-02-07
feat: allow to override standard cloud init paths

mudler:
This is needed in order to avoid loops

### DIFF
--- a/main.go
+++ b/main.go
@@ -667,6 +667,10 @@ The validate command expects a configuration file as its only argument. Local fi
 				Name:  "cloud-init-paths",
 				Usage: "Extra paths to add to the run stage",
 			},
+			&cli.StringSliceFlag{
+				Name:  "override-cloud-init-paths",
+				Usage: "Paths to use when running the stage",
+			},
 			&cli.BoolFlag{
 				Name:    "analyze",
 				Usage:   "Only print the modules that would run in the order they would run",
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -687,9 +691,14 @@ The validate command expects a configuration file as its only argument. Local fi
 			config, err := agentConfig.Scan(collector.Directories(constants.GetYipConfigDirs()...), collector.NoLogs)
 			config.Strict = c.Bool("strict")
 
+			if len(c.StringSlice("override-cloud-init-paths")) > 0 {
```
Itxaka, line 694: should this come after the extend?

main.go
```diff
@@ -667,6 +667,10 @@ The validate command expects a configuration file as its only argument. Local fi
 				Name:  "cloud-init-paths",
 				Usage: "Extra paths to add to the run stage",
 			},
+			&cli.StringSliceFlag{
+				Name:  "override-cloud-init-paths",
+				Usage: "Paths to use when running the stage",
```
Itxaka, line 672: Would expand the info to say that this just removes all default paths no?

Itxaka (approved):

main.go
mudler: I thought at it, but then would be more confusing : you could specify both but only one of the two would take effect.

mudler: mmm but this is the collector.. so things could come back from config :facepalm:

PR #694, opened 2025-03-04
Disable xattrs on rsync

Itxaka:
For systems with selinux enabled, this makes impossible to generate the proper files as its being blocked

This fixes it as tested on my Fedora with selinux.

Fixes #https://github.com/kairos-io/AuroraBoot/issues/196

### DIFF
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -185,7 +185,7 @@ func SyncData(log sdkTypes.KairosLogger, runner v1.Runner, fs v1.FS, source stri
 		"--human-readable",
 		"--archive", // recursive, symbolic links, permissions, owner, group, modification times, device files, special files
 		"--acls",    // preserve ACLS and permissions
-		"-U",        // preserve access times
+		"--atimes",  // preserve access times
 	}
 
 	for _, e := range excludes {
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -182,12 +182,9 @@ func SyncData(log sdkTypes.KairosLogger, runner v1.Runner, fs v1.FS, source stri
 		"--progress",
 		"--partial",
 		"--human-readable",
-		"--archive",
-		"--acls",
-		"-t",
-		"-U",
-		"-o",
-		"-g",
+		"--archive", // recursive, symbolic links, permissions, owner, group, modification times, device files, special files
+		"--acls",    // preserve ACLS and permissions
+		"-U",        // preserve access times
 	}
 
 	for _, e := range excludes {
--- a/pkg/utils/common.go
+++ b/pkg/utils/common.go
@@ -178,7 +178,17 @@ func SyncData(log sdkTypes.KairosLogger, runner v1.Runner, fs v1.FS, source stri
 	}
 
 	log.Infof("Starting rsync...")
-	args := []string{"--progress", "--partial", "--human-readable", "--archive", "--xattrs", "--acls"}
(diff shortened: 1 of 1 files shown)


pkg/utils/common.go
```diff
@@ -178,7 +178,17 @@ func SyncData(log sdkTypes.KairosLogger, runner v1.Runner, fs v1.FS, source stri
...
+	args := []string{
+		"--progress",
+		"--partial",
+		"--human-readable",
+		"--archive",
+		"--acls",
+		"-t",
+		"-U",
+		"-o",
+		"-g",
```
jimmykarily, line 190: Better use the long version of these flags.

jimmykarily:
I don't know much about selinux but `xattrs` is supposed to maintain "SELinux contexts" or "labels". I suppose we don't need it for as long as selinux is not supported right? Otherwise those metadata should be maintained?

pkg/utils/common.go
Itxaka: well, I was gonna put comments on what the flags do in the line after testing that all seems to work, will that work instead of long flags?

Itxaka:
> I don't know much about selinux but `xattrs` is supposed to maintain "SELinux contexts" or "labels". I suppose we don't need it for as long as selinux is not supported right? Otherwise those metadata should be maintained?

I think the issue here is, when building or rsyncing stuff in a host that has selinux, it autosets those selinux contexts to the extracted data, and then on copying those attrs cannot be maintained.

For building images, it will always fail when rsyncing stuff into a FAT partition as its impossible to save thos attributes in FAT.

Also, the question is, are those attrs the ones originally from the extracted OCI artifact, or are they new and set by  the host on data extraction? If they are set on data extraction, due to the host that extracts it having selinux, then that data is useless, as its new and not appropiate to the final system. AFAIK we dont save any selinux content on OCI artifact creation so thats why this only happens to fail in selinux hosts that build images.

IMHO, this is only a workaround for now. When we start dealing with selinux we will need to rework this properly to maintain the original selinux context if any and to deal with hosts that have selinux enabled and try to label things when we extract it :(

Itxaka:
> > I don't know much about selinux but `xattrs` is supposed to maintain "SELinux contexts" or "labels". I suppose we don't need it for as long as selinux is not supported right? Otherwise those metadata should be maintained?
> 
> I think the issue here is, when building or rsyncing stuff in a host that has selinux, it autosets those selinux contexts to the extracted data, and then on copying those attrs cannot be maintained.
> 
> For building images, it will always fail when rsyncing stuff into a FAT partition as its impossible to save thos attributes in FAT.
> 
> Also, the question is, are those attrs the ones originally from the extracted OCI artifact, or are they new and set by the host on data extraction? If they are set on data extraction, due to the host that extracts it having selinux, then that data is useless, as its new and not appropiate to the final system. AFAIK we dont save any selinux content on OCI artifact creation so thats why this only happens to fail in selinux hosts that build images.
> 
> IMHO, this is only a workaround for now. When we start dealing with selinux we will need to rework this properly to maintain the original selinux context if any and to deal with hosts that have selinux enabled and try to label things when we extract it :(

Ther eis also the possibility of having this wrong in the context of, you dont need to keep the original selinux labels as on first boot it should relabel everything.
I think that the original labels migth not be useful, as the selinux workflow is to relabel everything on first boot, but I have no idea if thats true and how that works with our immutable system....

jimmykarily (approved):

PR #702, opened 2025-03-12
Fix hooks under encrypted partitions

Itxaka:
We had a mess of mounting and unmounting things around when we try to copy things to persistent.

Part of the changes (using the by-label to mount the persistent) are due to the change in kcrypt. As we set the same label to the encrypted fs and unencrypted fs, our utils.Mount could get mistaken and return the first hit, which usually its the encrypted one, and we cannot mount that one.

This patch brings it up to date.

 - Makes bundles and logs hooks work when we have encrypted persistent. It didnt work before.
 - Makes both workflows the same.
 - Locks everything once its over, to not leave encrypted parts around
 - Mounts OEM so kcrypt can read the config if we are using a remote server for encryption
 - Mounts by label so there is not a change of getting the wrong device
 - Uses the mount syscall directly. The util can mistake and return the actual encrypted part if they both have the same label and finds it first

### DIFF
--- a/internal/agent/hooks/bundles.go
+++ b/internal/agent/hooks/bundles.go
@@ -60,13 +60,14 @@ func (b BundlePostInstall) Run(c config.Config, _ v1.Spec) error {
 				// doesnt seem to affect anything as the device is closed as expected so we ignore it if it matches the
 				// output of the error
 				if err != nil && !strings.Contains(out, "incorrect semaphore state") {
-					c.Logger.Errorf("could not close /dev/disk/by-label/%s: %s", p, out)
+					c.Logger.Warnf("could not close /dev/disk/by-label/%s: %s", p, out)
 				}
 			}
 		}()
 	}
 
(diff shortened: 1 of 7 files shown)


internal/agent/hooks/bundles.go
```diff
@@ -28,13 +34,55 @@ func (b BundlePostInstall) Run(c config.Config, _ v1.Spec) error {
...
 
-	err := os.MkdirAll("/usr/local/.state/var-lib-extensions.bind", os.ModeDir|os.ModePerm)
+	// Path if we have encrypted persistent and we are not on UKI
+	if len(c.Install.Encrypt) != 0 {
+		err := kcrypt.UnlockAll(false)
+		if err != nil {
+			return err
+		}
+	}
+	if len(c.Install.Encrypt) != 0 {
```
jimmykarily, line 51: This is the same condition as the one above. Merge them?

jimmykarily (approved):

jimmykarily (approved):

PR #713, opened 2025-03-19
Change webui assets path to be relative

mauromorales:
This allows me to do some complex config with my local server. For example if I spin up a vm and it gets the ip 192.168.122.137 within the same host I can simply access the webui from 192.168.122.137:8080 but let's say that I want to access it though a vpn like tailscale then I do something like this http://zeno/192.168.122.137/ which gets proxied to http://192.168.122.137:8080 which works fine, except for the assets because when the page gets rendered it tries to get them from http://zeno/node_modules

I'll also switch this to be done with a build in the future instead of accessing from node_modules directly

### DIFF
--- a/internal/webui/public/index.html
+++ b/internal/webui/public/index.html
@@ -6,25 +6,25 @@
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
 
   <!-- Bootstrap CSS -->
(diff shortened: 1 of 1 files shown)


jimmykarily (approved):

mauromorales:
need to figure out how to do this for the websocket :(

PR #738, opened 2025-04-03
First iteration of the sysext command

Itxaka:
Currently working:
 - list installed sysexts
 - list enabled sysext per active or passive
 - enable a installed sysext into active or passive
 - disable a installed sysext into active or passive
 - remove a installed sysext (disables it from active, passive as well)
 - install a sysext

The whole thing has no idea about versions or anything as that would
require heavy inspection of the underlying image

Only raw image is currently support, I would hessitate to include sysext in DIR format support as its a bit uglier and complex, plus doesnt support verity and signing

Part of https://github.com/kairos-io/kairos/issues/3068


mudler:
is christmas already?

mudler (approved):

pkg/action/sysext.go
```diff
@@ -0,0 +1,356 @@
...
+
+// TODO: Check which extensions are running
+// TODO: On disable we should check if the extension is running and refresh systemd-sysext?
+// TODO: On enable, should we refresh systemd-sysext immediately? Only if the current boot state equals the bootstate we enabled it for?
+
+const (
+	sysextDir        = "/var/lib/kairos/extensions/"
+	sysextDirActive  = "/var/lib/kairos/extensions/active"
+	sysextDirPassive = "/var/lib/kairos/extensions/passive"
+	sysextEnabledDir = "/run/extensions/"
```
mudler, line 55: where is this used? I can't see it used anywhere in the changeset

pkg/action/sysext.go
```diff
@@ -0,0 +1,356 @@
...
+	// first check if the extension is installed
+	extension, err := GetSystemExtension(cfg, ext, "")
+	if err != nil {
+		return err
+	}
+
+	var targetDir string
+	switch bootState {
+	case "active":
+		targetDir = sysextDirActive
```
mudler, line 135: mmm shouldn't the symlink go to `/run/sysextensions`?  what am I missing here?

Itxaka: you are missing the whole thing :D

This only creates symlinks into /var/lib/kairos/extensions/active/passive to mark those extensions as "enabled" for a boot state.

Then its immucore task on bootup to check those and copy them into /run/extensions during initramfs (or even link them again? Can you do that?) so when systemd-sysext service is up it founds those.

This means that we dont have to deal with /var/lib/extensions in persistent between boots (if you enable X in active, it will appear also on passive as they share persistent) and can manage them separately in our own dirs.

pkg/action/sysext.go
Itxaka: this is where immucore will copy the extensions on boot. I wondered if we were gonna use it in the future or not and had it here as a reference. Will remove as it seems we will not use it from here

pkg/action/sysext.go
Itxaka: There, now you can pass a `--now` flag to the enable command and if the extension you are enabling for a bootstate matches the current bootstate (so you are enabling X extension for active and you are on active) it will symlink it to /run/extensions and refresh the extensions, so it can be used inmediatly

mudler: ok cool - wasn't entirely clear neither from the PR/changeset that another piece of this its required. Any reason to use symlinks at this point? wouldn't be more clean to have a statefile that immucore reads?

Itxaka: I thought so at the start but it would end up basically as a database in a file. Symlinks are universal, you don't need specialized tools to create them or anything else, just a shell and it requires no deps to manage them. Plus systemd docs mention that it possible you should use symlinks to add syexts to the sysext directories so it felt natural to use those directly. Not much stuff that can go wrong with them, they are either there and valid or invalid and that's basically it, allows to maybe grow easily for specific boot states that we may want in the future and so on 

Having a statefile presented more issues that that in the future I think. Maybe not, I just wanted to make it as simple and accesible as possible

Itxaka: This way, immucore just read the dir a cording to the current boot then just goes over each entry enabling stuff.

That's it. No more to it.

Itxaka: Also think about this, you could have a different disk where you store your extensions and manually symlink them. done, simple, no need for other tools. Could even do so on cloud-init stuff before sysext is run. No need for the code that checks that statefile to be in both places, just simple file managing, solved problem for 3000 years :P

main.go
```diff
@@ -805,6 +806,221 @@ The validate command expects a configuration file as its only argument. Local fi
...
+						Name:  "active",
+						Usage: "Disable the system extension for the active boot entry",
+					},
+					&cli.BoolFlag{
+						Name:  "passive",
+						Usage: "Disable the system extension for the passive boot entry",
+					},
+				},
+				Before: func(c *cli.Context) error {
+					if c.Bool("active") && c.Bool("passive") {
```
jimmykarily, line 941: Why no `--now` in this one?

pkg/action/sysext.go
```diff
@@ -0,0 +1,384 @@
...
+// We cannot soft link it as EFI partition is FAT
+// To add/remove/update we need to deal with the actual files
+// Currently we are NOT using the measuring of the sysextensions so maybe we could move into a unified system with the same behaviour as above?
+// The sysextensions in Trusted boot are SIGNED so the measurement is not that important and immucore will refuse to move
+// them into /run/extensions if the signature doesnt match, so the measurements are not that important currently and
+// maybe they are not useful if sysext is the way of upgrading your system
+
+// So potentially we have a unified system where we have the same behaviour for both, but that measn that
+// anything provided by sysext will NOT be available until the initramfs stage where we enable the sysext service
+// we can potentially move the sysext service as early as possible once we mount the persistent dir
```
jimmykarily, line 40: I think we should just decide and update the comments to reflect the final decision. If measurement doesn't buy us anything, let's unify the 2 "modes". Since the directory where systemd expects the extensions in order to load them is always the same (trusted boot or not), I guess the reason to use symlinks in the non-trusted boot mode is to avoid copying files (for speed). Is that correct?

pkg/action/sysext.go
```diff
@@ -0,0 +1,384 @@
...
+// GetSystemExtension returns the system extension for a given name
+func GetSystemExtension(cfg *config.Config, name, bootState string) (SysExtension, error) {
+	// Get a list of all installed system extensions
+	installed, err := ListSystemExtensions(cfg, bootState)
+	if err != nil {
+		return SysExtension{}, err
+	}
+	// Check if the extension is installed
+	// regex against the name
+	re, err := regexp.Compile(name)
```
jimmykarily, line 108: [minor] we may need to constrain this a bit more to avoid things like "fips" matching "non-fips" or similar.

jimmykarily (approved):
Is this code only handling the non-trusted boot case? I only see code managing symlinks and such.

Itxaka:
> Is this code only handling the non-trusted boot case? I only see code managing symlinks and such.

ideally it should do both, as we have persistent on both cases, it should be able to deal with both of those. 

Only thing missing in there is the image policy, but that its set by the config files during initramfs so....

pkg/action/sysext.go
Itxaka: well its a regex no? So doing `^fips` should not match `non-fips`

pkg/action/sysext.go
Itxaka: yes, this comments should go away it was thefirst part only when I was starting

main.go
Itxaka: didnt had time xD

Itxaka:
and clarification on management part:

image

Itxaka:
The missing part as well is the cloud-config files that set the default values for sysext:
we already have the uki ones,w e need for non-uki: https://github.com/kairos-io/packages/blob/main/packages/static/kairos-overlay-files/files/system/oem/99_sysext.yaml

Itxaka:
Now fully working

PR #760, opened 2025-04-15
Use grub binaries and libs from rootfs

Itxaka:
We should not default to use the underlying OS files and binaries for
grub installs as we may be on a takeover scenario in which the
underlying OS does not provide the needed files.

This moves the grub binary and files to be provided by the source rootfs
and moves the grubOptions hook to use the go libraries directly instead
of shelling out to another binary

This also move the grub hook to use our existing go code instead of shelling out to grub2-editenv and moves the hook into the finish hooks, as it writes into OEM, on an encrypted disk it may not have access to the unencrypted partition, thus not being able to write to the real OEM.

This also brings a small fix in which we could be overwriting the existing vars as we didnt read the grub vars file before writing it, so there could potentially be a shipped file in there that we were fully overwriting.

Fixes https://github.com/kairos-io/kairos/issues/3330

### DIFF
--- a/internal/agent/hooks/gruboptions.go
+++ b/internal/agent/hooks/gruboptions.go
@@ -56,7 +56,7 @@ func grubOptions(c config.Config, opts map[string]string) error {
 			return err
 		}
 	}
-	err = utils.SetPersistentVariables(filepath.Join(runtime.OEM.MountPoint, "grubenv"), opts, c.Fs)
+	err = utils.SetPersistentVariables(filepath.Join(runtime.OEM.MountPoint, "grubenv"), opts, &c)
 	if err != nil {
 		c.Logger.Logger.Error().Err(err).Msg("Failed to set grub options")
 	}
--- a/internal/agent/hooks/gruboptions.go
+++ b/internal/agent/hooks/gruboptions.go
@@ -1,11 +1,13 @@
 package hook
 
 import (
+	"github.com/kairos-io/kairos-agent/v2/pkg/config"
+	cnst "github.com/kairos-io/kairos-agent/v2/pkg/constants"
 	v1 "github.com/kairos-io/kairos-agent/v2/pkg/types/v1"
-	"strings"
-
-	config "github.com/kairos-io/kairos-agent/v2/pkg/config"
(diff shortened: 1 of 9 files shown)


internal/agent/hooks/gruboptions.go
```diff
@@ -31,10 +33,32 @@ func (b GrubPostInstallOptions) Run(c config.Config, _ v1.Spec) error {
...
+		return err
 	}
-	c.Logger.Logger.Debug().Msg("Running GrubOptions hook")
+	c.Logger.Logger.Debug().Msg("Finish GrubOptions hook")
 	return nil
 }
+
+// grubOptions sets the grub options in the grubenv file
+// It mounts the OEM partition if not already mounted
+// If its mounted but RO, it remounts it as RW
```
jimmykarily, line 47: I don't see this anywhere in the code below (checking if it's mounted RO)

jimmykarily (approved):
can't say if it will fix the hetzner issue but code looks like it does what the description says.

internal/agent/hooks/gruboptions.go
Itxaka: `mounts.PrepareWrite` will do that for us, if its mounted and in the same mountpoint it will do nothing, if not mounted it will mount it and if mounted but RO it will remount it RW.

jimmykarily: but you only call `mounts.PrepareWrite` inside `if !runtime.OEM.Mounted {`. So if it's mounted, `PrepareWrite` is not called at all?

Itxaka:
Mark blocked until 3.4.0 is released

internal/agent/hooks/gruboptions.go
Itxaka: yes, because we by default mount oem as RW always, if its already mounted we assume its already RW. maybe we should change that?

I dont know, I dont like that this hook has this special mount thing while the rest have a strigth syscall.mount IIRC so Ill change this anyway to fall in line wiht others. ITs just that its what it was used already so I did not bother going deeper :D

PR #801, opened 2025-06-04
Fix the hooks order, unmounting and such

Itxaka:
Fixes UKI hooks which were untested since changing the grub ones.


mudler (approved):
Looking good

PR #833, opened 2025-07-01
Add registry flag to list-releases

Itxaka:
As we no longer set the repo and org for the releases, we allow to pass the registry flag to list the releases. Defaults to quay.io/kairos to maintain backwards compatibility

Fixes: https://github.com/kairos-io/kairos/issues/3504

### DIFF
--- a/main.go
+++ b/main.go
@@ -105,13 +105,14 @@ See https://kairos.io/docs/upgrade/manual/ for documentation.
 				Description: `List all available releases versions`,
 				Before: func(c *cli.Context) error {
 					// Check if the registry is set in the OS release and warn that its deprecated
-					_, err := sdkUtils.OSRelease("REGISTRY_AND_ORG")
+					registryAndOrg, err := sdkUtils.OSRelease("REGISTRY_AND_ORG")
 					if err == nil {
 						fmt.Println("Warning: The 'REGISTRY_AND_ORG' OS release variable is deprecated. Use the '--registry' flag instead.")
+						fmt.Println("Warning: Using the values from 'REGISTRY_AND_ORG' instead of the flag.")
+						_ = c.Set("registry", registryAndOrg)
+
 					}
-					if c.String("registry") == "quay.io/kairos" {
-						fmt.Println("Using default registry: quay.io/kairos")
-					}
+					fmt.Println(fmt.Sprintf("Using registry: %s", c.String("registry")))
 					return nil
 				},
 				Action: func(c *cli.Context) error {


main.go
```diff
@@ -98,9 +99,21 @@ See https://kairos.io/docs/upgrade/manual/ for documentation.
...
 				},
 				Name:        "list-releases",
 				Description: `List all available releases versions`,
+				Before: func(c *cli.Context) error {
+					// Check if the registry is set in the OS release and warn that its deprecated
+					_, err := sdkUtils.OSRelease("REGISTRY_AND_ORG")
+					if err == nil {
+						fmt.Println("Warning: The 'REGISTRY_AND_ORG' OS release variable is deprecated. Use the '--registry' flag instead.")
+					}
+					if c.String("registry") == "quay.io/kairos" {
```
jimmykarily, line 112: if `sdkUtils.OSRelease("REGISTRY_AND_ORG")` is set, do we want to be backwards compatible and use it (with a warning)? After all, we just want to make sure the command works when the env variable is not there. If it is, we are on a legacy system (somehow) so we can just use it.

This condition here, when true, it means they've used the new flag to set the registry to `quay.io/kairos`. Do we need to tell them we are using the default registry in this case? They've set it explicitly, we are not using the default, we are using what they set it to (which just happens to be the same as the default).

jimmykarily (approved):

main.go
Itxaka: TBH is there a way that the newer agent with this will be used in an older kairos release? I think not, so we can be safe with this and assume that any image built with this agent on it, will NOT have that variable as we control the init <-> agent relationship.

So this is only for images in which the creator is specifically setting that key in the kairos-release manually somehow, so I think its safe to just warn and continue

jimmykarily: > No, this is the default value of the flag, I wanted to make sure its printed so consumers see the registry being used when not passing the flag. If they set the flag to any other value, is meaningless to print it, its only when somebody does a list-releases with no flag, that the message will appear to inform them.

It also appears when you explicitly pass `--registry=quay.io/kairos` which is what most people are supposed to do anyway. I guess there is no easy way to tell if this is the default value or the user provided the same explicitly.

jimmykarily: I'm not sure either about the env variable. On one hand, it's easy to support it. On the other hand we will need to also make it optional input to `kairos-init` and we keep too many loose ends all over the place.

mauromorales (approved):

jimmykarily (approved):

PR #835, opened 2025-07-01
Deprecate `uri` in favor of `source`

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/3148


pkg/config/spec.go
```diff
@@ -1074,9 +1074,46 @@ func unmarshallFullSpec(r *Config, subkey string, sp v1.Spec) error {
...
 }
 
+// checkDeprecatedURIUsage checks if any Image structs in the spec have the deprecated URI field set
+// and logs a warning if found, also converts URI to Source for backwards compatibility
+func checkDeprecatedURIUsage(logger types.KairosLogger, sp v1.Spec) {
+	switch s := sp.(type) {
+	case *v1.InstallSpec:
+		checkImageURI(&s.Active, logger, "install.system")
+		checkImageURI(&s.Recovery, logger, "install.recovery-system")
+		checkImageURI(&s.Passive, logger, "install.passive")
```
jimmykarily, line 1090: Are all these even used? Need to check.

jimmykarily: removed the passive ones (we don't explicitly install or upgrade the passive)

pkg/config/spec.go
```diff
@@ -1074,9 +1074,43 @@ func unmarshallFullSpec(r *Config, subkey string, sp v1.Spec) error {
...
+		checkImageURI(&s.Active, logger, "install-uki.system")
+	case *v1.UpgradeUkiSpec:
+		checkImageURI(&s.Active, logger, "upgrade-uki.system")
+	}
+}
+
+// checkImageURI checks if an Image has the deprecated URI field set and handles the conversion
+func checkImageURI(img *v1.Image, logger types.KairosLogger, fieldPath string) {
+	if img.URI != "" {
+		logger.Warnf("The 'uri' field in %s is deprecated, please use 'source' instead", fieldPath)
```
jimmykarily, line 1105: this is what it looks like in the logs:

```
2025-07-02T07:18:22Z INF Boot in uki mode result=false
2025-07-02T07:18:22Z WRN [3939] The 'uri' field in upgrade.system is deprecated, please use 'source' instead
2025-07-02T07:18:25Z INF [3939] Setting image size to 3183MB
2025-07-02T07:18:25Z INF [3939] Checking if OCI image quay.io/kairos/alpine:3.21-standard-amd64-generic-v3.4.2-k3sv1.32.3-k3s1 exists
```

PR #846, opened 2025-07-21
Add grub edit var command

Itxaka:
Adds a subcommand to list and read vars from a grubenv file from the agent, which removes any need for grub2-editenv to be in the lvie image

Fixes https://github.com/kairos-io/kairos/issues/154

### DIFF
--- a/main.go
+++ b/main.go
@@ -819,6 +819,94 @@ The validate command expects a configuration file as its only argument. Local fi
 		},
 	},
 	{
+		Name:  "grubedit",
+		Usage: "grubedit subcommands",
+		Subcommands: []*cli.Command{
+			{
+				Name:      "list",
+				Usage:     "List all the grub variables from a file",
+				UsageText: "grubedit list FILE",
+				Before: func(c *cli.Context) error {
+					if c.Args().Len() != 1 {
+						cli.HelpPrinter(c.App.Writer, "File argument missing\n\n", c.Command)
+						_ = cli.ShowSubcommandHelp(c)
+						return fmt.Errorf("")
+					}
+					if _, err := os.Stat(c.Args().First()); os.IsNotExist(err) {
+						return fmt.Errorf("file %s does not exist", c.Args().First())
+					}
+					return nil
+				},
+				Action: func(c *cli.Context) error {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -818,6 +818,104 @@ The validate command expects a configuration file as its only argument. Local fi
...
+					return nil
+				},
+				Action: func(c *cli.Context) error {
+					config, err := agentConfig.Scan(collector.Directories(constants.GetUserConfigDirs()...), collector.NoLogs, collector.StrictValidation(c.Bool("strict-validation")))
+					if err != nil {
+						return err
+					}
+
+					// Transform the key=value argument into a map
+					parts := strings.SplitN(c.Args().First(), "=", 2)
```
jimmykarily, line 882: Where is the [padding](https://github.com/kairos-io/kairos-agent/pull/846/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R856) removed?

pkg/constants/constants.go
```diff
@@ -27,6 +27,7 @@ import (
 const (
 	GrubConf                     = "/etc/cos/grub.cfg"
 	GrubOEMEnv                   = "grub_oem_env"
+	GrubEnv                      = "grubenv" // grubenv is a file used by grub to store environment variables. Goes into OEM so its runtime modifiable
```
jimmykarily, line 30: What is the difference with `grub_oem_env` above? Maybe document the above file too, to make it clear what each one is used for?

jimmykarily (approved):

pkg/constants/constants.go
Itxaka: grub_oem_env seems to be used only for the install part, and it goes in the state folder, so it should be RO, while this one goes into OEM so its RW during runtime.

I have no idea why we would have 2, we may need to revisit this at some point to merge them? The grub_oem_env is not documented at all so it seems like its use its internal to tweak the grub behaviour ourselves or something? Unclear to me

main.go
Itxaka: Im not sure what does that means. The padding you linked is only for listing the grub variables for the user, for show basically so they look nice, but we dont add any padding to them when we manage them.

PR #940, opened 2025-08-26
fix: improve manual install argument handling

Itxaka:
Refactor the configuration generation for manual install CLI arguments to only include options that are explicitly set to true. This prevents unnecessary defaults from being included and returns an empty string if no options are provided.


internal/agent/install.go
```diff
@@ -380,15 +381,20 @@ func generateInstallConfForCLIArgs(sourceImageURL string) string {
...
+	}
 	if device != "" {
-		cfg += fmt.Sprintf(`
-  device: %s
-`, device)
+		cfg += fmt.Sprintf("  device: %s\n", device)
+	}
+	// if no options were passed, return empty string
+	if !reboot && !poweroff && device == "" {
+		return ""
```
jimmykarily, line 397: do this at the top to return early?

PR #944, opened 2025-08-27
feat: enhance installation process with error handling

Itxaka:
Improve the interactive installer by adding error handling and logging capabilities. This includes a dedicated error message display and adjustments to the installation flow to ensure proper user feedback during the installation process. Additionally, the logging mechanism has been refined to capture and report errors more effectively.

Fixes: https://github.com/kairos-io/kairos/issues/3605

### DIFF
--- a/internal/agent/TUIinstallProcessPage.go
+++ b/internal/agent/TUIinstallProcessPage.go
@@ -4,23 +4,31 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"github.com/kairos-io/kairos-sdk/types"
 	"os/exec"
 	"strings"
+	"sync"
 	"time"
+
+	"github.com/kairos-io/kairos-sdk/types"
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/lipgloss"
 )
 
+var internalLogger = types.NewKairosLogger("interactive-installer", "debug", true)
+
 // Install Process Page
 type installProcessPage struct {
-	progress int
-	step     string
-	steps    []string
-	done     chan bool   // Channel to signal when installation is complete
-	output   chan string // Channel to receive output from the installer
-	cmd      *exec.Cmd   // Reference to the running installer command
+	progress      int
+	step          string
+	steps         []string
+	done          chan bool
+	output        chan string
+	logsDone      chan bool
+	installerDone chan bool // Signal when installer is finished
+	once          sync.Once // Ensure goroutines are started only once
+	cmd           *exec.Cmd
+	errorMsg      string
 }
 
 func newInstallProcessPage() *installProcessPage {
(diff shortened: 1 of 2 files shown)


jimmykarily:
At some point, maybe we can implement something like this: https://youtu.be/2Xnvcqrt0Rc?t=365 (a toggle-able pane that shows logs).

internal/agent/TUIinstallProcessPage.go
```diff
@@ -38,84 +46,148 @@ func newInstallProcessPage() *installProcessPage {
...
-	logBuffer := bytes.Buffer{}
-	bufferLog := types.NewBufferLogger(&logBuffer)
-	cc.Logger = bufferLog
-
-	// Start the installer in a goroutine
-	go func() {
-		defer close(p.done)
-		err := RunInstall(cc)
-		if err != nil {
+	p.once.Do(func() {
```
jimmykarily, line 57: Why is `once` needed here? Are we running `Init()` multiple times?

internal/agent/TUIinstallProcessPage.go
```diff
@@ -38,84 +46,148 @@ func newInstallProcessPage() *installProcessPage {
...
+		internalLogger.Logger.Info().Msg("[Init] Starting installer and log goroutines")
+		oldLog := mainModel.log
```
jimmykarily, line 59: why is this called "old"? Where does the old logger print its messages? I see we print more things to internalLogger. We should use names that reflect the purpose of each logger. E.g. debugLogger and <whatever-the-old-logger-does>Logger 

As far as I can tell, there is no way to see the logs while installation is happening. What is the suggested user flow for monitoring the logs?

jimmykarily (commented):
I'm a bit confused on how the logs are accessed, so for now just some clarifications needed.

jimmykarily:
> > At some point, maybe we can implement something like this: [youtu.be/2Xnvcqrt0Rc?t=365](https://youtu.be/2Xnvcqrt0Rc?t=365) (a toggle-able pane that shows logs).
> 
> yes, I loved that log screen, its just that Im stupid around TUIs so Im having a hard time even doing the simplest installer ever....
> 
> Maybe you could take it over that add that at some point?

I can give it a shot, although I did that with Cursor's help. I only knew what I wanted it to look like :D Let's do it in another PR.

internal/agent/TUIinstallProcessPage.go
Itxaka: oldLog is the main config log which logs to file and journalctl. We cant use that one to track the install, so we store it and crete a new one that logs to a buffer so we can go over the buffer lines and know at which point of the install we are and if there is errors. The old logger is used to forward the lines we process from the new buffered logger so the logs end up in the journal as well, otherwise we would lose those logs.

There is a comment around there about using either a dual logger or something similar but thats not implemented as we need both a log that we can go over and to store the logs in persistent storage.

Sucks but this was like this before, it was a quick workaround for the current need.

jimmykarily: We can send them to `info` log level. At the end of the day, if something doesn't work, it's still the user that will collect logs and send them over to us. So, even if they don't read and understand the logs, they still might care to collect those. I'd say, it's better to stick to one logger for everything.

Itxaka:
@jimmykarily dropped the logger as it was only for my own debugging

PR #1179, opened 2026-04-05
feat: add phone-home feature

mudler:


internal/agent/phonehome_service.go
```diff
@@ -0,0 +1,95 @@
...
+		cmd := exec.Command(args[0], args[1:]...)
+		if out, err := cmd.CombinedOutput(); err != nil {
+			fmt.Printf("Warning: %s failed: %v (%s)\n", strings.Join(args, " "), err, string(out))
+		}
+	}
+}
+
+// hasPhonehomeConfig checks if any cloud-config file in the given directories
+// contains a `phonehome:` section with a `url` field.
+func hasPhonehomeConfig(dirs []string) bool {
```
jimmykarily: I suspect this should not be needed at all. "phone home" config should be treated as regular kairos config and we already have the Collector package to collect kairos config from the various files. Right before we call [`enablePhoneHomeIfConfigured`](https://github.com/kairos-io/kairos-agent/pull/1179/changes#diff-f48445f706f26bd59d7bae605c4445d3c178237d2489223bf9c1a813417d9713R68) we have already called the Collector so if we let it parse the phone-home config, we won't need this function here at all. I can try to refactor this if you want.

internal/phonehome/client_test.go
```diff
@@ -0,0 +1,342 @@
...
+			client := newTestClient("test-token")
+			err = client.Register(context.Background())
+			Expect(err).ToNot(HaveOccurred())
+
+			ms.mu.Lock()
+			defer ms.mu.Unlock()
+			Expect(ms.regCalls).To(Equal(0)) // no registration call made
+		})
+
+		It("should return error on invalid token", func() {
```
jimmykarily, line 228: We need one more test, to make sure that the token is also verified on Connection and not only registration. In other words, in the [Describe context below](https://github.com/kairos-io/kairos-agent/pull/1179/changes#diff-e94fc88fe49d04a0d644fca1766b56d1f90facff862d7542da6ee2f17f7f05cfR220), we should add a test that Registers with the correct token but tries to connect with an invalid one. That should fail.

internal/phonehome/handlers.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+// DefaultCommandHandler returns a CommandHandler that handles all daedalus commands.
+// The daedalusURL and apiKey are needed to download artifact images for upgrades.
+func DefaultCommandHandler(daedalusURL string, apiKey func() string) CommandHandler {
+	return func(cmd CommandData) (string, error) {
+		ctx := context.Background()
+
+		switch cmd.Command {
+		case "exec":
```
jimmykarily, line 37: We should consider making arbitrary command execution and "opt-in" (e.g. using a kairos config key). Maybe this should also include other actions that could destroy a Node (e.g. reset).

Even if this setup is meant to be local, at some point the machines might reach production. If the phone-home service is sill running, a simple DNS attack could redirect the nodes to a fake server and allow the attacker to run arbitrary commands on them.

We need to come up with clear instructions on how this setup is meant to be used and the steps needed before moving to production (e.g. pinned server certificates? disabling the phone-home service altogether?)

mudler: as of it works now it's entirely opt-in. IF the cloud config doesn't specify a `phonehome` stanza, no services will be installed in the box. 

But agree, it requires documentation, a way to filter out allowed commands and pinning certificates for self-signed certificates makes totally sense, but calls for a follow-up (I don't want to put more meat on this PR which is already too big for my eyes :D )

internal/phonehome/handlers.go
```diff
@@ -0,0 +1,262 @@
...
+		source = "oci:" + source
+	}
+
+	args := []string{"upgrade", "--source", source}
+	if cmd.Command == "upgrade-recovery" || cmd.Args["recovery"] == "true" {
+		args = append(args, "--recovery")
+	}
+
+	// Use background context — upgrade must NOT be killed if WS disconnects
+	fmt.Printf("[phonehome] Running: kairos-agent %s\n", strings.Join(args, " "))
```
Itxaka: do we need prints? we had several issues with prints breaking stuff already and this means not logging standard. Can we have a default logger here please so we can capture the logs form this phgonohome instead of having them around?

internal/phonehome/sysinfo.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+var (
+	cachedInfo     map[string]string
+	cachedInfoOnce sync.Once
+)
+
+// gatherSystemInfo collects OS, kernel, CPU, and memory info.
+// Results are cached since system info doesn't change between heartbeats.
+func gatherSystemInfo() map[string]string {
```
Itxaka, line 19: this seems really useful as to have it system wide utils instead of phonehome only.

internal/phonehome/sysinfo.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		cachedInfo = info
+	})
+	return cachedInfo
+}
+
+// mergeEnvFile parses KEY=VALUE lines (optional quotes) and merges into dst.
+// path is always a compile-time constant in this package (see gatherSystemInfo).
+func mergeEnvFile(dst map[string]string, path string) {
```
Itxaka, line 66: we have a lib in deps for key=value parsing no? godotenv is in there already, not sure if we need this merge stuff instead of relaying onto it for battel tested stuff?

main.go
```diff
@@ -604,6 +606,10 @@ This command is meant to be used from the boot GRUB menu, but can likely be used
 				Name:  "reset-oem",
 				Usage: "Reset the OEM partition. Warning: this will delete any persistent data on the OEM partition.",
 			},
+			&cli.BoolFlag{
```
Itxaka: Do we want to expose this? By defautl we agreed to always reset persistent, otherwise this will cause issues with leftovers in a supposed resetted system. What good is a reset if it doesnt really clean up?

Not sure I agree with exposing this here in the middle of a different PR to be honest.

Itxaka (commented):
Only 2 important issues and in fact only 1 really important:
 - reset-persistent flag exposure which I think needs to have it sown discussion and not be part of a drive-by commit in an unrelated PR.

The other important one is the logger, I dont think we need to go too much into this but we had issues with this before so if possible we should try to avoid printing to stdout. Not a blocked thought, just a heads up. Even a new logger with the phonehome name would be good enough IMHO.

main.go
mudler: but it was not the case, right? I could not see a "reset persistent" by default, but it was only an option enabled by events: https://github.com/kairos-io/kairos-agent/blob/05d10635dd034233aed9ccf9ff67599e55466bf8/internal/agent/reset.go#L166

This is just making it more explicit from the agent itself (without delegation), but we would need a handle to force resets with persistent partition data removed as well ( e.g. drop a new cloud config setup, reset, and act as a pristine node)

Itxaka: by default we had the persistent always resetted: https://github.com/kairos-io/kairos-agent/blob/main/pkg/config/spec.go#L542

we discussed this a long time ago (when redoing reset to be more streamlined) and decided that it was not good to expose this to the user IIRC as then the reset was not really doing anything if they did not removed persistent and could lead to weird things (no config as oem was removed but files and services still int here and whatnot)

IMHO, the exposed flag and whatnot should have been dropped as disabling persistent reset during a reset should NOT be a feature.

Itxaka (approved):

main.go
mudler: missed that - sorry - I thought we were NOT doing that by default, so that's why I've added it in the first place. Reverted in https://github.com/kairos-io/kairos-agent/pull/1179/commits/82eb23dc7fa8661a93722c2865c5ddb5384da013

mudler:
> Only 2 important issues and in fact only 1 really important:
> 
>     * reset-persistent flag exposure which I think needs to have it sown discussion and not be part of a drive-by commit in an unrelated PR.
> 
> 
> The other important one is the logger, I dont think we need to go too much into this but we had issues with this before so if possible we should try to avoid printing to stdout. Not a blocked thought, just a heads up. Even a new logger with the phonehome name would be good enough IMHO.

both point addressed!

- https://github.com/kairos-io/kairos-agent/pull/1179/commits/82eb23dc7fa8661a93722c2865c5ddb5384da013
- https://github.com/kairos-io/kairos-agent/pull/1179/commits/d7d0118c62d583db5b03150031c994bc707b94a2

PR #1188, opened 2026-04-14
Enhance --select to support partial matches for boot entries

### LINKED ISSUE #4038: kairos-agent bootentry --select broken on systemd < 257 in kairos 4.0.3 and kairos-init 0.8.5
`kairos-agent bootentry --select registration` fails on systems running systemd 255 (Ubuntu 24.04). The entry `registration` does not exist.

**System info:**
```
kairos-agent v2.27.0
kairos 4.0.3 and kairos-init 0.8.5
systemd 255 (255.4-1ubuntu8.14)
Ubuntu 24.04.4 LTS
```

**What happens:**

```
root@edge:~# kairos-agent bootentry --select registration
2026-04-11T05:32:53Z INF Kairos Agent version=v2.27.0
2026-04-11T05:32:53Z INF Boot Mode boot_mode=active_boot
2026-04-11T05:32:53Z INF Boot in uki mode result=true
2026-04-11T05:32:53Z INF Setting default boot entry to registration
2026-04-11T05:32:53Z ERR entry registration does not exist
```

The actual entries on the system are:
```
root@edge:~# ls /efi/EFI/kairos/
active.efi
passive.efi
recovery_install-mode_stylus.registration.efi
statereset_install-mode_stylus.registration.efi
```

There is no entry named `registration`. The entry that boots into registration mode is `recovery` 

**Impact:**

Our `80_stylus.yaml` cloud-init stages call `kairos-agent bootentry --select registration` in `after-reset` to set the node into registration mode after a reset. This silently fails, leaving the node stuck — it boots back into active with no config and crash-loops.

**Confirmed with older agent:**

On the another system with using kairos-agent v.3.5.9, kairos-init v0.5.28, kairos-agent v2.24.10, the reset flow works correctly — the node enters registration mode.

Itxaka:
When kairos-init >= 0.8.5 generates UKI boot entries with extended names (e.g. recovery_install-mode_stylus.registration.conf), the short form used by cloud-init scripts (--select registration) fails with "entry does not exist" because only exact matches were attempted.

After the exact-match check fails, fall back to substring matching against the original (pre-simplified) entry names. If exactly one entry matches, resolve and use it. If multiple entries match, return a clear "ambiguous" error instead of the misleading "does not exist".

Fixes: https://github.com/kairos-io/kairos/issues/4038

### DIFF
--- a/pkg/action/bootentries.go
+++ b/pkg/action/bootentries.go
@@ -156,7 +156,26 @@ func selectBootEntrySystemd(cfg *sdkConfig.Config, entry string) error {
 	err = entryInList(cfg, entry, entries)
 	// we also accept "active" as a selection so we can migrate eventually from cos
 	if err != nil && !strings.HasPrefix(entry, "active") {
-		return err
+		// Exact match failed against the (possibly simplified) entries list.
+		// Try a partial/substring match against the original entry names so that
+		// callers can use short names like "registration" to select an entry such
+		// as "recovery install-mode_stylus.registration".
+		var partialMatch string
+		for _, e := range originalEntries {
+			if strings.Contains(e, entry) {
+				if partialMatch != "" {
+					cfg.Logger.Errorf("entry %s is ambiguous, matches both %s and %s", entry, partialMatch, e)
+					return fmt.Errorf("entry %s is ambiguous", entry)
+				}
+				partialMatch = e
+			}
+		}
+		if partialMatch == "" {
+			return err
+		}
+		cfg.Logger.Debugf("entry %s matched %s via partial match", entry, partialMatch)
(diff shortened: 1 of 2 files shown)


mudler (approved):
makes sense. I'm just not sure what the proper UX should be here, if gating this with a specific flag (e.g. `--relaxed`, or the inverse `--exact-match`) to avoid to be ambigous

Itxaka:
I dont think this is even rigth, the entry is the first part, it always has been, this seems to be introducing random naming as part of the cmdline, which is set on all the entries w=hen using cmdline, so this doesnt gain aus anything

I will wait for this, this seems something else

PR #1190, opened 2026-04-17
fix(phonehome): address PR review feedback and gosec findings

mudler:
Route phonehome cloud-config through the shared Collector via a new LoadFromCollector helper, dropping the duplicate filepath.Walk in internal/agent/phonehome_service.go and the inlined parsing in main.go.

Add Config.AllowedCommands with a conservative default set (upgrade, upgrade-recovery, reboot); DefaultCommandHandler now fails-closed and requires operators to opt destructive commands (exec, reset, apply-cloud-config) in explicitly. This narrows the blast radius of a DNS-hijack / rogue-server scenario.

Add a client_test.go spec that registers with a valid token but connects with a rejected API key, exercising the WebSocket auth path.

Sweep gosec findings: tighten /oem dir perms (0755→0750) and OEM cloud-config write perms (0644→0600), validate artifact IDs before building /tmp paths and image URLs, explicitly discard best-effort errors on shutdown paths, and annotate the remaining false positives (filesystem paths flagged as secrets, fixed-argument exec.Command calls) with //nosec + rationale.

### DIFF
--- a/internal/phonehome/client_test.go
+++ b/internal/phonehome/client_test.go
@@ -24,6 +24,7 @@ type mockServer struct {
 	mu       sync.Mutex
 	regCalls int
 	lastReg  phonehome.RegisterRequest
+	wsToken  string
 	// WS tracking
 	wsConnected    chan struct{}
 	heartbeats     []phonehome.HeartbeatData
@@ -31,9 +32,22 @@ type mockServer struct {
 	commandsToSend []phonehome.CommandData
 }
 
+func (ms *mockServer) setWSToken(t string) {
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
+	ms.wsToken = t
+}
+
+func (ms *mockServer) getWSToken() string {
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
+	return ms.wsToken
+}
+
 func newMockServer(validToken string) *mockServer {
 	ms := &mockServer{
 		wsConnected: make(chan struct{}, 1),
+		wsToken:     "test-api-key",
 	}
 
 	mux := http.NewServeMux()
@@ -69,7 +83,7 @@ func newMockServer(validToken string) *mockServer {
 
 	mux.HandleFunc("/api/v1/ws", func(w http.ResponseWriter, r *http.Request) {
 		token := r.URL.Query().Get("token")
-		if token != "test-api-key" {
+		if token != ms.getWSToken() {
 			http.Error(w, "unauthorized", http.StatusUnauthorized)
 			return
 		}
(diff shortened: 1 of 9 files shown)


internal/phonehome/client_test.go
```diff
@@ -259,6 +273,25 @@ var _ = Describe("PhoneHome Client", func() {
 			Expect(len(ms.heartbeats)).To(BeNumerically(">=", 2))
 		})
 
+		It("should fail when the stored API key is rejected by the server", func() {
+			// Register with a valid registration token so credentials are stored,
+			// then make the server reject the resulting API key on the WS handshake.
+			// This covers the case where the registration token is correct but the
+			// node's API key is invalid/revoked at connection time.
+			client := newTestClient("test-token")
```
jimmykarily, line 281: Duplicated in line 159. Better store it in a var and use the var here (e.g. validTestToken)

internal/phonehome/config.go
```diff
@@ -20,12 +26,71 @@ type Config struct {
...
+	// When nil, defaults to DefaultAllowedCommands (upgrade/reboot only).
+	// Destructive commands (exec, reset, apply-cloud-config) must be opted-in
+	// explicitly because a DNS hijack / rogue server can reach this endpoint.
+	AllowedCommands []string `yaml:"allowed_commands"`
+}
+
+// DefaultAllowedCommands is the conservative set of commands a phone-home node
+// will execute when the user has not specified allowed_commands in cloud-config.
+// It intentionally excludes exec, reset and apply-cloud-config.
+var DefaultAllowedCommands = []string{"upgrade", "upgrade-recovery", "reboot"}
```
jimmykarily, line 39: `upgrade` can also be used to hack a box but upgrading to a tampered image. This can be avoided with trusted boot though so I'd say it's ok to leave it here. In any case, we should create a followup ticket to:
- Document proper protection against possible attacks (e.g. use trusted boot mode)
- Implement pinned certificate for the server so that a DNS hijack is not enough to fool the Nodes into connecting to a malicious instance.

jimmykarily: Or we can add a "allowed registry and org" kind of thing. We can check if the upgrade source comes from the allowed registry and org when upgrading. Again, that's a follow-up ticket.

internal/phonehome/handlers.go
```diff
@@ -164,19 +189,45 @@ func writeOEMCloudConfig(content string) error {
...
+func isSafeArtifactID(id string) bool {
+	if id == "" {
+		return false
+	}
+	for _, r := range id {
+		switch {
+		case r >= 'a' && r <= 'z':
+		case r >= 'A' && r <= 'Z':
+		case r >= '0' && r <= '9':
+		case r == '-' || r == '_' || r == '.':
```
jimmykarily, line 227: Could the dot be a problem as well? (e.g. `/tmp/..` could expand to `/` , depending how it's used in the code).

jimmykarily (approved):

PR #1230, opened 2026-06-04
fix(grub): skip shim copy on riscv64 EFI install

mauromorales:
## Summary

- Skip `copyShim()` on `riscv64` during EFI installation; Fedora/RHEL ship `grub2-efi-riscv64` but no shim package for this architecture.
- Write `BOOTRISCV64.EFI` as the UEFI removable-media fallback when copying `grubriscv64.efi` (same approach as AuroraBoot ISO builder).
- Add unit test covering riscv64 EFI install without shim.

Fixes manual installation failure:

```
could not find any shim file to copy
```

reported while installing a Fedora-based riscv64 Kairos image.

## Context

`kairos-init` already installs `grub2-efi-riscv64` / `grub2-efi-riscv64-modules` and documents booting grub directly (no shim). AuroraBoot was updated in kairos-io/AuroraBoot#509; this aligns `kairos-agent` installer behavior.

Tracking: kairos-io/kairos#4083

## Test plan

- [x] Unit test: `installs with efi firmware on riscv64 without shim`
- [x] Manual: install Fedora riscv64 Kairos live ISO to disk via EFI; confirm install completes and system boots from installed ESP

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -148,6 +148,8 @@ func GetFallBackEfi(arch string) string {
 	switch arch {
 	case ArchArm64:
 		return "bootaa64.efi"
+	case ArchRiscv64:
+		return "BOOTRISCV64.EFI"
 	default:
 		return "bootx64.efi"
 	}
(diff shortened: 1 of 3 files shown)


pkg/constants/constants.go
Itxaka, line 147: noice, we probably need to set a link here in case this keeps growing to the spec https://uefi.org/specs/UEFI/2.10/03_Boot_Manager.html#removable-media-boot-behavior

Itxaka (approved):

PR #1236, opened 2026-06-09
feat(tui): add prereqs sanity-check contract and bus wiring

### LINKED ISSUE #4102: SPIKE: Installation hooks to surface custom alerts during interactive install
## Context

Slack: https://spectrocloud.slack.com/archives/C056VJ4V1HP/p1779912784031779

The interactive TUI installer (`kairos-agent` TUI) currently has a fixed set of steps that display during installation:

- Preparing installation
- Partitioning disk
- Running before-install
- Installing Active
- Configuring bootloader
- Creating Recovery
- Creating Passive
- Running after-install

The TUI process page (`TUIinstallProcessPage.go`) monitors log output and maps specific log messages to step progress. However, there's no mechanism for custom hooks to surface arbitrary messages, warnings, or alerts to the user during the interactive install.

Related: #4101 - Cleanup of leftover data on additional disks and prompt to wipe before installer handoff

## Problem Statement

When installing interactively, the installer cannot surface custom alerts or warnings to the user during the installation process. For example:
- Warning about leftover data on additional disks that will be wiped
- Custom branding messages
- Provider-specific alerts
- Any pre-installation validations that need user acknowledgment

The current hook system (`pkg/action/install.go`) runs `before-install` and `after-install` hooks, but these run silently in the background and their output isn't surfaced to the TUI.

## Spike Goals

- [ ] Investigate the current hook execution flow and where hooks run relative to the TUI
- [ ] Explore options for surfacing hook output to the TUI:
  - Custom hook types that return messages to the TUI
  - A "custom step" mechanism in the provider/plugin system
  - Log-based message passing (parse hook output and display in TUI)
  - Cloud-init stage hooks that run during interactive install
- [ ] Evaluate the impact on UKI vs non-UKI install paths
- [ ] Consider UX implications: blocking vs non-blocking alerts, dismissable vs confirmation-required
- [ ] Draft a proposal for the approach

## Relevant Files

- `kairos-agent/internal/agent/TUIinstallProcessPage.go` - TUI install process page, parses logs to track steps
- `kairos-agent/internal/agent/TUIinstallOptionsPage.go` - TUI install options page
- `kairos-agent/internal/agent/install.go` - RunInstall entry point
- `kairos-agent/internal/agent/hooks/hook.go` - Hook interface and hook lists (FinishInstall, PostInstall, etc.)
- `kairos-agent/pkg/action/install.go` - InstallAction.Run() which orchestrates the install and calls hooks
- `kairos-agent/pkg/uki/install.go` - UKI install path
- `kairos-agent/pkg/constants/` - Hook constant definitions (BeforeInstallHook, AfterInstallHook, etc.)

## Considerations

- Both non-UKI and UKI install paths need to be considered
- The TUI currently polls log output to determine progress - any new hook mechanism should integrate with this
- Hooks should be configurable per provider or globally
- Need to balance between flexibility and complexity
- The existing `before-install` and `after-install` cloud-init stages already exist - can we leverage those?

Itxaka:
Introduce pkg/prereqs defining the contract for interactive-install
pre-installation sanity checks: the EventChecks / EventChecksApply bus
events, the Check / Decision / payload structs, and pure helpers
(ParseChecks, Blocker, BuildDecisions). These mirror the existing
EventInteractiveInstall customization mechanism.

Wire the two new events into the agent plugin Manager via
internal/bus.AllEvents so provider plugins are subscribed to them.

Provider check failures are reported via Check data (SeverityError),
not the go-pluggable EventResponse.Error field, to avoid the bus's
global os.Exit-on-error handler killing the TUI.

Fixes https://github.com/kairos-io/kairos/issues/4102

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/tui-check-examples/tui-check-diskwipe/main.go
+++ b/tui-check-examples/tui-check-diskwipe/main.go
@@ -1,0 +1,181 @@
+// tui-check-diskwipe is an example Kairos interactive-install check plugin
+// (a "tui-check"). It participates in the installer's pre-installation sanity
+// checks, a plugin set separate from the generic agent providers.
+//
+// It demonstrates the EventChecks / EventChecksApply round-trip defined in
+// pkg/prereqs:
+//
+//   - On EventChecks it scans the system's disks and, for any disk that still
+//     carries Kairos partition labels (COS_*) from a previous installation, it
+//     returns a multi-select check letting the user pick which disks to wipe.
+//   - On EventChecksApply it wipes the disks the user selected in the TUI
+//     (wipefs) and reports the result.
+//
+// Build it and drop the resulting binary, named with the "tui-check-" prefix,
+// into one of the check-plugin search paths (/system/tui-checks,
+// /usr/local/system/tui-checks, or the agent's working directory):
+//
+//	go build -o tui-check-diskwipe ./tui-check-examples/tui-check-diskwipe
+//	install -m0755 tui-check-diskwipe /usr/local/system/tui-checks/
+//
+// Anything this plugin writes to stdout/stderr is captured by go-pluggable into
+// the EventResponse.Logs field and re-logged by the agent, so the log lines
+// below show up in the agent's journal/log.
+//
+// NOTE: wiping a disk is destructive. The action only runs for disks the user
+// explicitly selects on the prerequisites screen.
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/kairos-io/kairos-agent/v2/pkg/prereqs"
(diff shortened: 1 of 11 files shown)


tui-check-examples/tui-check-diskwipe/main.go
```diff
@@ -0,0 +1,181 @@
+// tui-check-diskwipe is an example Kairos interactive-install check plugin
```
mudler, line 1: I just wonder if we should have rather a canonical places for examples, e.g. `examples/sanity-checks`, etc.

Itxaka: I dont think it matters, we havent needed an examples up until now so...

PR #1237, opened 2026-06-09
fix(tui): paginate disk selection list when it overflows the terminal

### LINKED ISSUE #4128: [TUI] Disk selection page doesn't paginate - cursor goes out of view when many disks are present
## Problem

The disk selection page in the kairos interactive installer TUI does not implement pagination when the number of disks exceeds the visible terminal area. When users navigate through a long list of disks using arrow keys, the cursor (left `>` arrow) moves past the visible area and becomes inaccessible to the user.

### Reproduction Steps
1. Boot the kairos interactive installer
2. Select a machine with many disks (e.g., 10+ disks as shown in the screenshot)
3. Use `↓` or `j` to navigate down the list
4. Observe that the `>` cursor moves beyond the visible TUI area and is no longer visible

### Current Behavior
From `internal/agent/TUIdiskPage.go`, the `View()` method renders all disks in a flat loop without any height constraint:

```go
func (p *diskSelectionPage) View() string {
    s := "Select target disk for installation:\n\n"
    s += "WARNING: All data on the selected disk will be DESTROYED!\n\n"

    for i, disk := range p.disks {
        cursor := " "
        if p.cursor == i {
            cursor = lipgloss.NewStyle().Foreground(kairosAccent).Render(">")
        }
        s += fmt.Sprintf("%s %s (%s)\n", cursor, disk.name, disk.size)
    }

    return s
}
```

There is no logic to:
- Calculate available height for the disk list
- Slice the disk list to only show items within the visible area
- Track scroll offset and update cursor position

### Expected Behavior
- The disk list should be constrained to the available vertical space
- When navigating past the visible items, the list should scroll (new items appear at the top, old items disappear from the bottom)
- The cursor should always remain visible within the displayed portion of the list

## Proposed Solution

The fix should be applied in `kairos-agent/internal/agent/TUIdiskPage.go`. The approach should follow patterns used elsewhere in the TUI (e.g., `TUImodel.go` has `effectiveSize()` for calculating available dimensions).

### Action Items

1. **Calculate available height** for the disk list in the `View()` method, accounting for:
   - Header text ("Select target disk for installation:")
   - Warning message
   - Help text at the bottom
   - Border padding

2. **Add a scroll offset field** to `diskSelectionPage`:
   ```go
   type diskSelectionPage struct {
       disks   []diskStruct
       cursor  int
       offset  int  // top visible item index
   }
   ```

3. **Implement scroll logic** in `Update()`:
   - When `cursor` moves past the visible window, adjust `offset` accordingly
   - Ensure `offset` never goes negative and never exceeds `len(disks) - visibleCount`

4. **Update `View()` to render only visible items**:
   - Slice `p.disks` based on `offset` and `visibleCount`
   - Adjust cursor display to account for the offset

5. **Add a scroll indicator** (optional but recommended):
   - Show `...` at the top or bottom of the list when items are hidden
   - This gives the user visual feedback that more items exist beyond the visible area

### Example Pseudocode for View()
```go
func (p *diskSelectionPage) View() string {
    // Calculate available height for the list
    availableHeight := height - headerLines - warningLines - helpLines - padding
    
    // Clamp offset and cursor
    if p.cursor < p.offset {
        p.offset = p.cursor
    }
    if p.offset > p.cursor {
        p.offset = p.cursor
    }
    if p.offset+availableHeight < len(p.disks) {
        p.offset = len(p.disks) - availableHeight
    }
    if p.offset < 0 {
        p.offset = 0
    }
    
    // Render only visible items
    start := p.offset
    end := min(start+availableHeight, len(p.disks))
    
    for i := start; i < end; i++ {
        // render disk i
    }
}
```

### References
- Similar pagination logic can be found in `TUImodel.go` (`effectiveSize()`) and other TUI pages that use `lipgloss` for layout
- The `lipgloss` library supports `Height()` and `Width()` constraints that could be leveraged

### Environment
- kairos-agent TUI uses `charmbracelet/bubbletea` and `charmbracelet/lipgloss`
- Affected file: `internal/agent/TUIdiskPage.go`

Itxaka:
The disk selection page rendered every disk in a flat loop with no height constraint. With many disks (10+) the cursor moved past the visible area and Model.View truncated the overflow, leaving the cursor inaccessible.

Add a scroll window: track an offset, compute how many rows fit (mirroring Model.View's height-10 budget minus the header lines), clamp the offset so the cursor is always visible, and render only the visible slice with "... more above/below" indicators when items are hidden.

Fixes https://github.com/kairos-io/kairos/issues/4128

### DIFF
--- a/internal/agent/TUIdiskPage.go
+++ b/internal/agent/TUIdiskPage.go
@@ -29,13 +29,17 @@ type diskSelectionPage struct {
 // + blank line.
 const diskPageHeaderLines = 4
 
+// visibleRows is the number of rows reserved by Model.View for page content
+// (height minus this value gives the usable content area).
+const visibleRows = 10
+
 // visibleCount returns how many disk rows fit in the available vertical space.
-// It mirrors the content budget applied by Model.View (height-10) and reserves
-// two lines for the scroll indicators when the list overflows.
+// It mirrors the content budget applied by Model.View (height-visibleRows) and
+// reserves two lines for the scroll indicators when the list overflows.
 func (p *diskSelectionPage) visibleCount() int {
 	_, height := effectiveSize(mainModel.width, mainModel.height)
-	// Model.View slices each page's content to height-10 lines.
-	avail := height - 10 - diskPageHeaderLines
+	// Model.View slices each page's content to height-visibleRows lines.
+	avail := height - visibleRows - diskPageHeaderLines
 	if len(p.disks) > avail {
 		// Reserve room for the top/bottom "..." scroll indicators.
 		avail -= 2
--- a/internal/agent/TUIdiskPage.go
+++ b/internal/agent/TUIdiskPage.go
@@ -21,6 +21,51 @@ type diskStruct struct {
 type diskSelectionPage struct {
 	disks  []diskStruct
 	cursor int
+	offset int // index of the first visible disk in the scroll window
+}
+
+// diskPageHeaderLines is the number of fixed lines the page renders before the
+// disk list: the "Select target disk" prompt + blank line, and the WARNING line
+// + blank line.
+const diskPageHeaderLines = 4
+
+// visibleCount returns how many disk rows fit in the available vertical space.
+// It mirrors the content budget applied by Model.View (height-10) and reserves
+// two lines for the scroll indicators when the list overflows.
+func (p *diskSelectionPage) visibleCount() int {
+	_, height := effectiveSize(mainModel.width, mainModel.height)
+	// Model.View slices each page's content to height-10 lines.
+	avail := height - 10 - diskPageHeaderLines
+	if len(p.disks) > avail {
+		// Reserve room for the top/bottom "..." scroll indicators.
+		avail -= 2
+	}
+	if avail < 1 {
+		avail = 1
+	}
+	return avail
+}
+
+// clampOffset keeps the scroll window so the cursor stays visible and the
+// offset never runs past the end of the list.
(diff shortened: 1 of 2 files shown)


internal/agent/TUIdiskPage.go
```diff
@@ -21,6 +21,51 @@ type diskStruct struct {
...
+// + blank line.
+const diskPageHeaderLines = 4
+
+// visibleCount returns how many disk rows fit in the available vertical space.
+// It mirrors the content budget applied by Model.View (height-10) and reserves
+// two lines for the scroll indicators when the list overflows.
+func (p *diskSelectionPage) visibleCount() int {
+	_, height := effectiveSize(mainModel.width, mainModel.height)
+	// Model.View slices each page's content to height-10 lines.
+	avail := height - 10 - diskPageHeaderLines
```
jimmykarily: What is the magic number `10`? What should it match? I don't understand what "content budget" means in the comment.

jimmykarily (approved):

internal/agent/TUIdiskPage.go
Itxaka: its basically 10 visible "rows" or lines, so 10 items at a time.

jimmykarily: Make it a constant then? (e.g. `VISIBLE_ROWS`)

Itxaka: @copilot Check the comment above and send a pr to fix this

PR #1240, opened 2026-06-10
feat: dispatch interactive install to external installer + progress markers

mudler:
Purely-refactoring, Agent-side groundwork for extracting the interactive installer into a separate, separately-owned kairos-installer binary.

- interactive-install becomes a dispatcher: it resolves an external installer (KAIROS_INSTALLER env -> kairos-installer on PATH) and execs it with the tty inherited, propagating its exit code; it falls back to the current in-process bubbletea TUI when none is found.
- New pkg/progress emits stable, machine-readable progress markers (KAIROS_INSTALL_STEP <step> / KAIROS_INSTALL_ERROR <msg>) on stdout, gated by KAIROS_AGENT_PROGRESS so human installs and the in-process TUI are unaffected. InstallAction.Run() emits one marker per major step.
- docs/installer-contract.md documents the stable contract


pkg/progress/progress.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// EventStep is the "event" value for a per-step progress line.
+	EventStep = "step"
+	// EventError is the "event" value for a failure line.
+	EventError = "error"
+)
+
+// Step identifiers — the fixed, documented enum of the contract.
+const (
+	StepPartition     = "partition"
```
jimmykarily: Maybe it would make sense to let the kairos-agent decide on the progress percentage by emitting that too. In other words have a specific event (e.g. `StepProgress` which emits values between 0 and 100).

Also we can let the rest of the steps be arbitrary to allow the kairos-agent to change without the consumer (the installer) needing to adapt. If in the future we decide to add or remove steps, or the steps take longer than before for whatever reason, we can change that in the kairos-agent and all the installers will adapt automatically becuase they will rely on the information provided by kairos-agent to show progress and steps.

In other words, don't write the steps or the progress in stone but rather let the kairos-agent report that back.

jimmykarily: For the steps part, we can just "un-expose" the constants here to prevent the installers from consuming them directly and expecting them to not change.

mudler: would make sense to have percentage applied to the step, so it can be re-emitted, e.g.:

- active, progress: 10%
- before-install, progress: 100, etc


but I think that makes sense as a follow-up. I wanted to keep the scope very lean here and be mostly a refactoring

jimmykarily: It's just that we are now setting the "contract". If people implement installers based on that contract, it will be harder to change it.

mudler: I'll move the contract directly to the SDK, so it needs to be imported and would break during compilation

PR #1254, opened 2026-06-15
feat: support pulling images from insecure registries

jimmykarily:
Wire the kairos-sdk WithInsecureRegistry support through kairos-agent so install and upgrade can pull container images from registries served over plain HTTP or presenting an untrusted/self-signed TLS certificate.

- Replace the local imageextractor.OCIImageExtractor with a type alias to the SDK's canonical image.OCIImageExtractor (single source of truth).
- Add an Insecure field to the install/upgrade specs (and their uki variants) plus a --insecure CLI flag on upgrade/install/manual-install. Both the flag and the install.insecure/upgrade.insecure cloud config keys flow through applyInsecureExtractor, which swaps the default extractor for {Insecure: true} before any pull or size calculation.
- Drop the crane.Manifest pre-flight in pkg/config/spec.go and rely on the SDK GetImage/GetOCIImageSize error, which is insecure-aware. This removes the dual-path smell (a second pull path that ignored insecure). The normal upgrade path already failed fast on a missing image via the fatal GetSourceSize error; the uki upgrade path swallowed that error, so make a container-source size error fatal there too to preserve fail-fast.

Co-Authored with Claude AI


Itxaka (commented):
shit, is there any other key name that we could use? insecure_registry maybe? I dont like the way of having install.insecure in there

I know this is technically bikeshedding but I really do think this very important so consumers do not panic on seeing this stuff.

Itxaka (approved):

Itxaka (approved):

PR #1265, opened 2026-06-18
Validate partitions fit disk

### LINKED ISSUE #2972: Check that manual partitions don't exceed the disk's size

jimmykarily:
Fail install early when manual partitions exceed the disk size

Closes kairos-io/kairos#2972

Problem

When a user defines manual partition sizes in the install cloud-config whose sum is larger than the target disk, the install used to proceed anyway and produce an unbootable system that only failed at first boot — e.g. immucore not finding COS_STATE, ending in a reboot loop (see the diagnosis on kairos-io/kairos#2950). There was no validation comparing the requested partition sizes against the target disk capacity, and the partitioner's flex-partition math (pkg/partitioner/disk.go) can even underflow when the fixed partitions already exceed the disk.

Change

InstallSpec.Sanitize() now sums the requested partition sizes (via PartitionsByInstallOrder, so firmware partitions are included) and fails with a clear error if they don't fit the target disk:
the requested partitions size (30001MB) does not fit in the target disk "/dev/vda" (20000MB), which also needs 2MB reserved for partition metadata

Details:
- Reserves the GPT metadata overhead the partitioner needs (1MB start alignment + 1MB backup GPT header), so a layout that would just barely not fit is rejected up front instead of failing mid-partitioning.
- Partitions with size: 0 ("take the remaining space") don't count toward the requested total.
- If the target disk can't be resolved yet (e.g. target unset), the check is skipped — no behavior change for existing paths.

Tests

- Unit (pkg/implementations/spec/spec_test.go, via the ghw mock): rejects an over-size layout, rejects a layout that fits the raw disk but leaves no room for GPT metadata, and accepts a fitting layout.
- e2e (tests/e2e/partition_validation_test.go): drives kairos-agent manual-install against the 20GB e2e VM disk — a 30000MB persistent config must fail at sanitization (before the image is pulled), and a fitting 1024MB persistent config must clear the check and reach the pull stage. Added the partition-validation label to the e2e label filter.

 Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -790,11 +790,12 @@ func NewUkiUpgradeSpec(cfg *sdkConfig.Config) (*spec.UpgradeUkiSpec, error) {
 		return spec, fmt.Errorf("could not read host partitions")
 	}
 
-	// Get free size of partition
+	// Get free size of partition in MiB so it can be compared with the image
+	// sizes, which are also in MiB.
 	var stat unix.Statfs_t
 	_ = unix.Statfs(spec.EfiPartition.MountPoint, &stat)
-	freeSize := stat.Bfree * uint64(stat.Bsize) / 1000 / 1000
-	cfg.Logger.Debugf("Partition on mountpoint %s has %dMb free", spec.EfiPartition.MountPoint, freeSize)
+	freeSize := stat.Bfree * uint64(stat.Bsize) / 1024 / 1024
+	cfg.Logger.Debugf("Partition on mountpoint %s has %dMiB free", spec.EfiPartition.MountPoint, freeSize)
 	// Check if the source is over the free size
 	if spec.Active.Size > uint(freeSize) {
 		return spec, fmt.Errorf("source size(%d) is bigger than the free space(%d) on the EFI partition(%s)", spec.Active.Size, freeSize, spec.EfiPartition.MountPoint)
--- a/pkg/config/spec.go
+++ b/pkg/config/spec.go
@@ -175,7 +175,7 @@ func NewInstallSpec(cfg *sdkConfig.Config) (*spec.InstallSpec, error) {
 	if err != nil {
 		cfg.Logger.Warnf("Failed to infer size for images: %s", err.Error())
 	} else {
-		cfg.Logger.Infof("Setting image size to %dMb", size)
+		cfg.Logger.Infof("Setting image size to %dMiB", size)
 		spec.Active.Size = uint(size)
 		spec.Passive.Size = uint(size)
 		spec.Recovery.Size = uint(size)
@@ -214,7 +214,7 @@ func NewInstallElementalPartitions(log sdkLogger.KairosLogger, spec *spec.Instal
 		MountPoint:      constants.OEMDir,
 		Flags:           []string{},
 	}
-	log.Infof("Setting OEM partition size to %dMb", oemSize)
+	log.Infof("Setting OEM partition size to %dMiB", oemSize)
 	// Double the space for recovery, as upgrades use the recovery partition to create the transition image for upgrades
 	// so we need twice the space to do a proper upgrade
 	// Check if the default/user provided values are enough to fit the images sizes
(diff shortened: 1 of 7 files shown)


pkg/implementations/spec/spec.go
```diff
@@ -101,7 +104,47 @@ func (i *InstallSpec) Sanitize() error {
...
+// not fit instead of letting the partitioning fail mid-install.
+const gptMetadataOverhead = 2 * 1024 * 1024
+
+// checkPartitionsFitTargetDisk returns an error if the sum of the requested
+// partition sizes does not fit in the target disk once the GPT metadata
+// overhead is reserved. A partition with size 0 means "take whatever space is
+// left on the disk", so it does not count towards the requested total. If the
+// target disk was not found (for example when the target is not set yet) the
+// check is skipped.
+func (i *InstallSpec) checkPartitionsFitTargetDisk(targetDisk *partitions.Disk) error {
```
Itxaka, line 130: are we sure this is all correct? AI tends to mess up sectors vs bytes vs bits and whatnot

Itxaka (approved):

pkg/implementations/spec/spec.go
jimmykarily: That's why I asked it to create an e2e test for it :D

jimmykarily: For what it's worth, I got it investigating:

The units check out — it's bytes-vs-bytes, 1024-based throughout, no sector/bit conversion anywhere in the path:

  - Partition.Size is MiB (kairos-sdk ghw/disk_partition_handler.go: Size: uint(size / (1024 * 1024))), and the partitioner reads it back the same way (pkg/partitioner/disk.go: uint64(part.Size * 1024 *
  1024)).
  - Disk.SizeBytes is bytes (kairos-sdk ghw/ghw.go: return size * sectorSize, sectors × 512).
  - So uint64(requestedMB)*1024*1024 + gptMetadataOverhead > targetDisk.SizeBytes compares bytes to bytes.

  Two deliberate choices worth calling out:
  1. The comparison is requested + overhead > size, not size - overhead, so a disk smaller than the overhead can't underflow the uint64.
  2. The 2 MiB overhead isn't arbitrary — it's exactly the partitioner's reservations: 1 MiB of start alignment (first partition starts at 1024 * 1024 / sectorSize) + 1 MiB for the backup GPT header.
  For a size: 0 ("rest of disk") layout the partitioner computes diskSize - (1MiB + fixedSum) - 1MiB, and this check is precisely the guard that keeps that subtraction non-negative — the original
  underflow that produced the unbootable systems.

  One honest caveat: the values are labelled "MB" in code/messages while they're really MiB, kept that way to match the existing convention (config/spec.go logs %dMb). Happy to switch to MiB if you
  prefer precision over consistency.

Itxaka: > One honest caveat: the values are labelled "MB" in code/messages while they're really MiB, kept that way to match the existing convention (config/spec.go logs %dMb). Happy to switch to MiB if you
prefer precision over consistency.

plz tell claude to do this so there is no longer any confusion

Itxaka (approved):

pkg/config/spec.go
```diff
@@ -790,11 +790,12 @@ func NewUkiUpgradeSpec(cfg *sdkConfig.Config) (*spec.UpgradeUkiSpec, error) {
...
 
-	// Get free size of partition
+	// Get free size of partition in MiB so it can be compared with the image
+	// sizes, which are also in MiB.
 	var stat unix.Statfs_t
 	_ = unix.Statfs(spec.EfiPartition.MountPoint, &stat)
-	freeSize := stat.Bfree * uint64(stat.Bsize) / 1000 / 1000
-	cfg.Logger.Debugf("Partition on mountpoint %s has %dMb free", spec.EfiPartition.MountPoint, freeSize)
+	freeSize := stat.Bfree * uint64(stat.Bsize) / 1024 / 1024
+	cfg.Logger.Debugf("Partition on mountpoint %s has %dMiB free", spec.EfiPartition.MountPoint, freeSize)
```
jimmykarily, line 798: I'm sure I remember we did a similar fix recently. One in which we changed this to be the correct `1024` but then we run out of "slack" and we had to increase some other value. I can't find it though. @Itxaka am I imagining things here?

jimmykarily: yup: https://github.com/kairos-io/AuroraBoot/commit/63320d6ea29ddac98ffd1ec2706cb76e1f8d2508

Updated the commit message to mention the previous history

Itxaka: Yes we hit annissue with Ubuntu 26 i think? Not sure the issues behind It but the quickest was ro icntease the size a bit. Pretty sure this all needs a rerwork at some point as we should be calcularing the size correctly with no error margin

PR #1266, opened 2026-06-18
fix: stream sysext files to avoid OOM on large extensions

Dr-N00B:
Replace buffering entire .sysext.raw files into memory with streaming via io.Copy(). This fixes OOMKilled crashes for large extensions (~500MB) on resource-constrained pods.

Changes:
- fileSource.Download() now streams files in chunks instead of buffering
- File permissions are preserved

Tests:
- Added "should preserve file permissions when streaming"
- Added "should correctly copy large files using streaming"


mudler (approved):

Itxaka (commented):
very nice improvement!

PR #1269, opened 2026-06-19
fix: run grub-install chrooted into source rootfs for non-kairos liveCDs

jimmykarily:
manual-install executed grub-install directly on the host, so its ELF interpreter and shared libraries resolved against the liveCD rather than the installation source. This worked only on a Kairos liveCD and failed on Ubuntu (missing interpreter) or Fedora (missing libdevmapper).

Run grub-install chrooted into the active mountpoint with chroot-relative paths and the boot directory bind-mounted, so it resolves everything from the source image and installs work from any liveCD.

This will allow users to download the kairos-agent on any livecd and perform an installation to a selected disk with a command like this:

```
sudo ./kairos-agent --debug manual-install --device /dev/vda --source oci://quay.io/kairos/hadron:v0.4.0-core-amd64-generic-v4.1.2 c.yaml
```

It is also useful in cases in which all we get is a rescue console (e.g. hetzner) and we can't boot ISOs and such (our take-over instructions).

### DIFF
--- a/pkg/utils/grub.go
+++ b/pkg/utils/grub.go
@@ -68,14 +68,10 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool, 
 
 		r, _ := state.NewRuntimeWithLogger(g.config.Logger.Logger)
 		if r.BootState == state.LiveCD {
-			// get the current dir
-			currentDir, err := os.Getwd()
-			if err != nil {
-				g.config.Logger.Warnf("Failed getting current dir: %s", err)
-				return err
-			}
-
-			liveOsDir := filepath.Join(currentDir, "LiveOS_rootfs")
+			// grub-install runs chrooted into rootDir (see below), so its working
+			// dir is the chroot root. Create the symlink under rootDir so it
+			// resolves to CWD/LiveOS_rootfs from inside the chroot.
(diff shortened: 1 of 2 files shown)


pkg/utils/grub.go
```diff
@@ -125,7 +111,35 @@ func (g Grub) Install(target, rootDir, bootDir, grubConf, tty string, efi bool,
...
+		// its original path.
+		chrootRoot := strings.TrimSuffix(rootDir, "/")
+		grubBinChroot := strings.TrimPrefix(grubBin, chrootRoot)
+		grubDirChroot := strings.TrimPrefix(grubdir, chrootRoot)
+
+		grubargs = append(
+			grubargs,
+			fmt.Sprintf("--directory=%s", grubDirChroot),
+			fmt.Sprintf("--boot-directory=%s", bootDir),
+			"--target=i386-pc", "-v",
```
mauromorales, line 134: why is the target i386-pc?

jimmykarily: Strangely, this means "legacy bios": https://wiki.archlinux.org/title/GRUB#Installation_2

This code is not new btw, it was just moved here.

mudler (approved):

mauromorales (approved):

PR #1270, opened 2026-06-19
ci: add codecov configuration

jimmykarily:
Coverage uploaded to Codecov comes only from the unit-tests workflow. A large part of this project is validated by end to end tests that run inside a VM and do not contribute to that coverage profile, so Codecov's strict defaults flag those e2e-only code paths as missing coverage on every pull request.

Add a codecov.yml that tolerates a small dip in project coverage, makes patch coverage informational instead of blocking, and ignores test scaffolding and generated mocks.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,36 @@
+# Codecov configuration.
+#
+# Coverage uploaded to Codecov comes only from the unit-tests workflow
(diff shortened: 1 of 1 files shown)


Itxaka (changes requested):
There is already a codecov config file.

